PR #7, opened 2022-01-07
update readme add SSL_TYPE

flx413:


Kulsk:
thanks for finding this problem, I'm quite sorry that I forget to update README. but your pull request is still with problem, would mind if I fix it?

flx413:
OK, my pull request closed.

PR #92, opened 2022-02-16
[=] add notice for usage of test_client

Kulsk:
connect failure to different servers with test_client and session ticket, transport parameters, token from different servers or hosts happens again, this is better to be noticed in docs/docs-zh/Testing-zh.md.


RuiqiZhou (approved):
LGTM

PR #109, opened 2022-03-04
[!] increase init_idle_time_out to 64 seconds

Kulsk:
just to increase the chance of successful handshake under high loss networks


RuiqiZhou (approved):
LGTM 👍

PR #112, opened 2022-03-06
[=] add specifications for interop demo

Kulsk:
the contents under original demo directory is always misunderstood, redefine it as interop


Saigut:
Hello, why not continue this pr? @Kulsk @ruiqizhou

PR #123, opened 2022-03-29
Update README-zh.md

robertsong2019:
since -j is not specified with number, it will exhausted build machine resource, and make it hard to execute other command, fix to 4


RuiqiZhou:
The QuickStart is just a suggestion, you can choose according to your own situation. We do not need to restrict.

RuiqiZhou:
The QuickStart is just a suggestion, you can choose according to your own situation. We do not need to restrict.

PR #136, opened 2022-04-06
[!] fix bug in xqc_str_hash_init when malloc return NULL

wangweiwei1188:
fix bug in xqc_str_hash_init when malloc  return NULL


RuiqiZhou (approved):
LGTM 👍

PR #137, opened 2022-04-06
[-]Remove function call which's not needed.

wangweiwei1188:
Remove function call which's not needed.


RuiqiZhou:
I think it would be better if you could briefly explain about why this function is not needed.

RuiqiZhou (approved):
LGTM 👍  Thanks

PR #143, opened 2022-04-13
[+] Add demo: simple io client/server

Saigut:
Add a simple io demo, which do simple read/write on stream, and will display transmit speed.


RuiqiZhou:
The original "demo" directory is actually used for interoperability testing, not demos. We are renaming it to "interop", see PR #112  for details. 
- #112

The PR for Issue #148  will be submitted shortly.
- #148 

So, my suggestion is not to put your simple_demo in the "demo" directory, but to create a new directory to avoid ambiguities and conflicts.

Saigut:
How about this file structure?
```
xquic/
|- interop/
|- demo/
    |- demo_simple_io_common.h
    |- demo_simple_io_common.c
    |- demo_simple_io_server.c
    |- demo_simple_io_client.c
...
```
Or this:
```
xquic/
|- interop/
|- demo-simple/
    |- demo_simple_io_*
...
```
I think the first file structure would be more reasonable.  What I would do is to add 4 new files into `demo/` directory, and PR #112 can create `interop/` and then move origin demo files to `interop/`.

RuiqiZhou:
> How about this file structure?
> 
> ```
> xquic/
> |- interop/
> |- demo/
>     |- demo_simple_io_common.h
>     |- demo_simple_io_common.c
>     |- demo_simple_io_server.c
>     |- demo_simple_io_client.c
> ...
> ```
> 
> Or this:
> 
> ```
> xquic/
> |- interop/
> |- demo-simple/
>     |- demo_simple_io_*
> ...
> ```
> 
> I think the first file structure would be more reasonable. What I would do is to add 4 new files into `demo/` directory, and PR #112 can create `interop/` and then move origin demo files to `interop/`.

I think it's reasonable.

@Kulsk But does this affect your current development work? I think you should review this PR and evaluate the impact.

Y7n05h:
Thank you very much. The demo you added helped me a lot to understand the usage of `xquic`. But is this a bug? My understanding is that `continue` should be executed here instead of `break`.

https://github.com/alibaba/xquic/blob/a5550ea6dd8091fa9f6a9129f2c8b1b2be83d6e6/demo/demo_simple_io_server.c#L465-L467

Saigut:
> Thank you very much. The demo you added helped me a lot to understand the usage of `xquic`. But is this a bug? My understanding is that `continue` should be executed here instead of `break`.
> 
> https://github.com/alibaba/xquic/blob/a5550ea6dd8091fa9f6a9129f2c8b1b2be83d6e6/demo/demo_simple_io_server.c#L465-L467

Hi, here should be `break`. Because here should not wait for data in callback.

Y7n05h:
> Hi, here should be `break`. Because here should not wait for data in callback.
Oh, yes, you are right.

Saigut:
Hello, since `interop` has been added, I think we can continue to consider merging this pr.

Kulsk:
> Hello, since `interop` has been added, I think we can continue to consider merging this pr.

What is this demo supposed to do? As there are already samples for stream read/write cases, it is not suggested to add another similar demo. I'm wondering if this io demo can be remade as a speed tester or some other tools else, to make it more different.

Kulsk (changes requested):
duplicated features are not suggested, maybe this one can be remade as a Tool

PR #256, opened 2022-10-29
Update xqc_id_hash.h

chunlei-wu:
Change the issue in xqc_pow2, when you run the inline function on Windows, in the while(), the n will be change. Adding the temp and using it in the while will be OK.


yuanbo-zhang:
Dear      您好，邮件我已收到 ！BestwishThink's张渊博

yuanbo-zhang:
xqc_pow2 can be remove，it can be replaced by the same function xqc_pow2_upper

PR #300, opened 2023-04-24
fix: stream offset overflow issue in 32-bit system

alexliu-liu:
Stream offset may be transfer from uint64_t to size_t when gen stream frame. In 32bit system, size_t means unsign int. This may cause stream offset error when offset larger than 2^32.


ruiqizhou (approved):
LGTM 👍 Thanks!

PR #306, opened 2023-05-10
[!] fix cid limit error

Kulsk:
cid limit comparison error makes xquic fails interop with those implementations which supports multiple cids


Kulsk:
fixed on alibaba/xquic interop branch directly

Issue #301, opened 2023-04-25
[Feature]: TLS证书压缩


kukuhua1:
### Before request a new feature

- [X] I have searched the [issues](https://github.com/alibaba/xquic/issues) of this repository and believe that this is not a duplicate.

### Describe the feature you'd like supported

xquic依赖了babassl,babassl支持了证书压缩功能，请问如何在xquic中启用证书压缩功能呢？

Kulsk:
is there any detailed information or interface of this feature?

kukuhua1:
xquic并没有调用tongsuo(babassl)提供的证书压缩api,但是xquic编译依赖了babassl库，是不是可以在源码中调用该接口，然后在测试程序中实现证书压缩呢？

PR #309, opened 2023-06-25
在XQUIC上实现TLS证书压缩

kukuhua1:
该PR用于解决 #301


Kulsk:
the code is not well designed and get no reply from author

PR #318, opened 2023-07-12
[=] add usage specifications for test_client and test_server

Kulsk:
fix usage function description on test_client and test_server


Yanmei-Liu (approved):
Looks good to me.

PR #360, opened 2023-11-03
Fix Bug in xqc_engine.c, Bug: Reinject unacked packet twice 

ShengtongZhu:
In xqc_conn_schedule_packets_to_paths, the unacked packets are reinjected. After that, they are reinjected again. So, those lines of code should be deleted.


yangfurong:
It is not a bug. Here, we try to reinject unacked packets **immediately** after they have been committed into the network. In `xqc_conn_schedule_packets_to_paths`, we can't do it as the packets are not sent yet (thus, they are not **unacked**).

PR #381, opened 2023-11-29
Fix bug: after skip_path, log message is wrong

ShengtongZhu:
After executing "skip_path", some variables of the log message do not belong to the current path. So, it should get the value first, then execute "skip_path".


ShengtongZhu (commented):
SKIP_PATH

ShengtongZhu:
SKIP path

PR #389, opened 2023-12-18
Allow building with dynamic libssl

iczero:
The cmake build will always statically link `libssl` and `libcrypto`. In some cases, it may be preferable to dynamically link them. This adds `-DSSL_DYNAMIC=1` which will build `libxquic` dynamically linked to `libssl` and `libcrypto` and also disable `libxquic-static`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,15 +60,17 @@ if(NOT SSL_INC_PATH AND NOT SSL_LIB_PATH)
     find_package(SSL)
 
     if(NOT SSL_FOUND)
-        message(FATAL_ERROR "ssl moudle not found")
+        message(FATAL_ERROR "ssl module not found")
     endif(NOT SSL_FOUND)
 
-    if(NOT SSL_LIBRARIES_STATIC)
-        message(FATAL_ERROR "ssl/crypto static library not found")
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -58,15 +58,17 @@ endif()
...
-    message(FATAL_ERROR "ssl/crypto static library not found")
-endif(NOT SSL_LIBRARIES_STATIC)
-
 # compat with elder macros
-set(SSL_LIB_PATH ${SSL_LIBRARIES_STATIC})
+if(SSL_DYNAMIC)
+    message(NOTICE "building with dynamic libssl")
+    set(SSL_LIB_PATH ${SSL_LIBRARIES})
+else()
+    message(NOTICE, "building with static libssl")
```
Kulsk, line 69: remove char ',' after NOTICE

iczero:
Rebased and fixed the format problem.

Please note the "not found" message is no longer required. If the library is not found, cmake will print the following:

```
-- Could NOT find SSL (missing: SSL_LIBRARY_STATIC CRYPTO_LIBRARY_STATIC) 
CMake Error at CMakeLists.txt:63 (message):
  ssl module not found
```

Kulsk (approved):
LGTM

PR #390, opened 2023-12-18
[=] allow user specify the path of ssl include and dependent library

Kulsk:


yuanbo-zhang (approved):
optimize SSL library dependencies

PR #410, opened 2024-04-12
small optimizations for XQUIC

Ya-Pasha-364shy:


Ya-Pasha-364shy:
Hello everyone. Here I have prepared a small request containing some optimizations for the xquic implementation, which, it seems to me, have a right to exist. I will be glad to receive feedback. 

- Best regards, Pavel

src/transport/xqc_multipath.h
```diff
@@ -88,6 +88,7 @@ typedef enum {
 } xqc_path_specified_flag_t;
 
 typedef enum {
+    XQC_PATH_CLASS_PERF_CLASS_INVALID = -1,
```
yangfurong, line 91: No need to define this.

src/transport/xqc_engine.c
```diff
@@ -439,36 +443,50 @@ xqc_engine_create(xqc_engine_type_t engine_type,
                                engine->config->cfg_log_level_name,
                                &engine->eng_callback.log_callbacks, engine->user_data);
     if (engine->log == NULL) {
+        XQC_ENGINE_ERROR_LOGGER("<%s:%d> unable to initialize logger in engine\n",
```
yangfurong, line 446: Remove these logging codes or use the xquic-style logging macro.

src/transport/scheduler/xqc_scheduler_minrtt.c
```diff
@@ -103,24 +98,20 @@ xqc_minrtt_scheduler_get_path(void *scheduler,
                 best_path[path_class] ? best_path[path_class]->path_id : -1);
     }
 
-
-    for (path_class = XQC_PATH_CLASS_AVAILABLE_HIGH; 
-         path_class < XQC_PATH_CLASS_PERF_CLASS_SIZE; 
-         path_class++)
+    if (best_path_class != XQC_PATH_CLASS_PERF_CLASS_INVALID
```
yangfurong, line 101: The original code intends to find the path with the minimal RTT for each path class and selects the one with the lowest path class index. You modification breaks that.

yangfurong (changes requested):

src/transport/scheduler/xqc_scheduler_minrtt.c
Ya-Pasha-364shy: ok, but I don’t understand why you need to choose the first path you come across with a minimum RTT value with a lowest path class index, if there is a possibility that further there will be a path with an even lower RTT, albeit with a bigger path class index

Ya-Pasha-364shy: thus, as long as this probability exists, `xqc_scheduler_minrtt.c` will return the path not with the smallest RTT

Ya-Pasha-364shy:
Hello !
I tried to correct all the comments, please look at it in your free time

src/transport/xqc_engine.c
```diff
@@ -437,38 +437,54 @@ xqc_engine_create(xqc_engine_type_t engine_type,
                                engine->config->cfg_log_event,
                                engine->config->cfg_log_timestamp,
                                engine->config->cfg_log_level_name,
-                               &engine->eng_callback.log_callbacks, engine->user_data);
+                               &engine->eng_callback.log_callbacks,
+                               engine->user_data);
+
     if (engine->log == NULL) {
+        xqc_log(engine->log, XQC_LOG_ERROR,
```
yangfurong, line 444: Referencing a NULL pointer.

src/transport/scheduler/xqc_scheduler_minrtt.c
```diff
@@ -7,6 +7,7 @@
 #include "src/transport/scheduler/xqc_scheduler_common.h"
 #include "src/transport/xqc_send_ctl.h"
 
+#define INVALID_PATH_CLASS -1
```
yangfurong, line 10: No need to define this macro.

src/transport/scheduler/xqc_scheduler_minrtt.c
```diff
@@ -103,24 +99,20 @@ xqc_minrtt_scheduler_get_path(void *scheduler,
                 best_path[path_class] ? best_path[path_class]->path_id : -1);
     }
 
-
-    for (path_class = XQC_PATH_CLASS_AVAILABLE_HIGH; 
-         path_class < XQC_PATH_CLASS_PERF_CLASS_SIZE; 
-         path_class++)
+    if (best_path_class != INVALID_PATH_CLASS
```
yangfurong, line 102: In the original implementation. The scheduler classifies paths into different classes according to their bandwidth/rtt/loss_rate/etc. The classes have different priorities and a lower index leads to a higher priority. For each class, we need to find the path whose RTT is the minimum in the class. At the end, we will select the minrtt path from the highest priority class. So, your code breaks our original logic. 

Indeed, the name of the scheduler, "minrtt", may not exactly describe what it does. It's a bit confusing.

yangfurong (changes requested):

src/transport/scheduler/xqc_scheduler_minrtt.c
Ya-Pasha-364shy: ok, thanks for the explanation about this path's classes. I will revert original version of this piece of code

src/transport/xqc_engine.c
```diff
@@ -437,38 +437,53 @@ xqc_engine_create(xqc_engine_type_t engine_type,
                                engine->config->cfg_log_event,
                                engine->config->cfg_log_timestamp,
                                engine->config->cfg_log_level_name,
-                               &engine->eng_callback.log_callbacks, engine->user_data);
+                               &engine->eng_callback.log_callbacks,
+                               engine->user_data);
+
     if (engine->log == NULL) {
+        fprintf(stderr, "unable to initialize logger in engine\n");
```
yangfurong, line 444: Please remove it.

yangfurong (changes requested):

yangfurong (changes requested):

tests/test_client.c
```diff
@@ -1512,13 +1508,6 @@ xqc_client_create_path_socket(xqc_user_path_t *path,
         return XQC_ERROR;
     }
 #ifndef XQC_SYS_WINDOWS
-    if (path_interface != NULL
-        && xqc_client_bind_to_interface(path->path_fd, path_interface) < 0)
```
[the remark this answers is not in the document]
Ya-Pasha-364shy: because the `xqc_client_bind_to_interface` function is already called in the `xqc_client_create_socket` function on line 1504 in this file.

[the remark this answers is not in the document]
Ya-Pasha-364shy: making another call that associates the returned socket descriptor with the same name (`path_interface`) is needlessly

yangfurong (approved):
LGTM

PR #412, opened 2024-04-22
[!] fix PTO error when HANDSHAKE_DONE is not sent;

Kulsk:


src/transport/xqc_cid.c
```diff
@@ -250,7 +250,7 @@ xqc_int_t
 xqc_cid_switch_to_next_state(xqc_cid_set_t *cid_set, xqc_cid_inner_t *cid, xqc_cid_state_t next_state)
 {
     if (xqc_cid_in_cid_set(cid_set, &cid->cid) == NULL) {
-        return -XQC_ECONN_CID_NOT_FOUND;
+        return XQC_OK;
```
Yanmei-Liu, line 253: Should not hide the error inside the function

Yanmei-Liu: About solve the issue with CID retirement retransmission, we need to check the state before switch the CID state, instead of letting the wrong state pass.

Kulsk: fixed by validating CID after receiving a RC frame

Yanmei-Liu (approved):
Looks good to me.

PR #417, opened 2024-05-05
[~] added token complexity for client-server interaction

### LINKED ISSUE #266: [Bug]: Address validation token is too easy.
### What happened?

> [8.1.4. ](https://www.rfc-editor.org/rfc/rfc9000.html#section-8.1.4)[Address Validation Token Integrity](https://www.rfc-editor.org/rfc/rfc9000.html#name-address-validation-token-in)
An address validation token MUST be difficult to guess. Including a random value with at least 128 bits of entropy in the token would be sufficient, but this depends on the server remembering the value it sends to clients.

Xquic use ip address + expire time as token.

### Steps To Reproduce

Read the code.

### Relevant log output

_No response_

Ya-Pasha-364shy:
Hello everyone, I decided to complicate the token by double-xor encryption with an initialization vector and a connection secret, this secret is unique for each connection. I will be glad to constructive criticism and reviews.

Fixes #266 issue

* Best regards, Pavel Chernov

### DIFF
--- a/src/transport/xqc_cid.c
+++ b/src/transport/xqc_cid.c
@@ -82,9 +82,11 @@ xqc_cid_set(xqc_cid_t *cid, const unsigned char *data, uint8_t len)
     }
 }
 
+static unsigned char g_token_buf[XQC_MAX_TOKEN_LEN];
 static unsigned char g_scid_buf[XQC_MAX_CID_LEN * 2 + 1];
 static unsigned char g_dcid_buf[XQC_MAX_CID_LEN * 2 + 1];
 static unsigned char g_sr_token_buf[XQC_STATELESS_RESET_TOKENLEN * 2 + 1];
+
 
 unsigned char *
 xqc_dcid_str(const xqc_cid_t *dcid)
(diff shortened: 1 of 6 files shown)


Ya-Pasha-364shy:
Hello everyone, I decided to complicate the token by double-xor encryption with an initialization vector and a connection secret, this secret is unique for each connection. I will be glad to constructive criticism and reviews

Ya-Pasha-364shy:
Hi all. Do you find these changes useful or can I close the pull request?

Yanmei-Liu:
> Hi all. Do you find these changes useful or can I close the pull request?

@Ya-Pasha-364shy Thanks for the PR, and sorry for the late reply. We find it useful but it can't be merged at the current state. There's an important issue here: The generated token need to be used during new connection handshaking in the future. While usually we used it in distributed cluster, and the servers in the cluster don't have to share state with each other. Then the current solution would cause the server can't validate the token in the future without current state.

We have discussed a new solution to solve this issue. Basically we'd like to have the modification upon your PR, and keep this PR unmerged until this new solution is merged into this branch.

Ya-Pasha-364shy:
> @Ya-Pasha-364shy Thanks for the PR, and sorry for the late reply. We find it useful but it can't be merged at the current state. There's an important issue here: The generated token need to be used during new connection handshaking in the future. While usually we used it in distributed cluster, and the servers in the cluster don't have to share state with each other. Then the current solution would cause the server can't validate the token in the future without current state.
> 
> We have discussed a new solution to solve this issue. Basically we'd like to have the modification upon your PR, and keep this PR unmerged until this new solution is merged into this branch.

Thx for feedback. I understand you, I will waiting for decision of checking new connection handshaking by token in the future. Tag me here when the new solution is merged into `main`.

PR #427, opened 2024-05-10
[!] fix handle Retire Prior To error when NEW_CONNECTION_ID arrives and retire all the previous CIDs

Kulsk:
if peer issues a NEW_CONNECTION_ID frame and retire all the previous CIDs with Retire Prior To field, XQUIC will consider there is not enough CID and fails to handle this case as XQUIC don't insert the new CID in its dcid_set before handle Retire Prior To field.

Insert CID before handling Retire Prior To field, and allow temporary exceed of local CID limit.


src/transport/xqc_frame.c
```diff
@@ -840,13 +819,54 @@ xqc_process_new_conn_id_frame(xqc_connection_t *conn, xqc_packet_in_t *packet_in
         return ret;
     }
 
-    ret = xqc_cid_set_insert_cid(&conn->dcid_set.cid_set, &new_conn_cid, XQC_CID_UNUSED, conn->local_settings.active_connection_id_limit);
+    /* An endpoint MAY send connection IDs that temporarily exceed a peer's 
+     * limit if the NEW_CONNECTION_ID frame also requires the retirement of any
+     * excess, by including a sufficiently large value in the Retire Prior To
+     * field. Hence it is not reasonable to consider it as an error */
+    cid_limit += retire_prior_to;
```
Yanmei-Liu, line 826: [The "Retire Prior To" field](https://datatracker.ietf.org/doc/html/rfc9000#name-new_connection_id-frames) is defined as an absolute value, not a relative value.

Kulsk: resolved by checking "(total_dcid_cnt - dcid_to_be_retired_cnt) > max_active_cid_limit"

PR #440, opened 2024-07-11
fix xqc_log_event parameter bug

dreamwind1985:
fix the vulnerability which has been found in XQUIC v1.8.0 from xquic-opensource。


Yanmei-Liu (approved):
Looks good.

PR #466, opened 2024-12-04
[+] Add support for multipath draft-11 and 12

Yanmei-Liu:
https://datatracker.ietf.org/doc/draft-ietf-quic-multipath/


demo/demo_client.c
```diff
@@ -1745,7 +1828,7 @@ xqc_demo_cli_parse_server_addr(char *url, xqc_demo_cli_net_config_t *cfg)
     /* set hint for hostname resolve */
     struct addrinfo hints = {0};
     memset(&hints, 0, sizeof(struct addrinfo));
-    hints.ai_family = AF_UNSPEC;    /* Allow IPv4 or IPv6 */
+    hints.ai_family = AF_INET; //AF_UNSPEC;    /* Allow IPv4 or IPv6 */
```
yangfurong, line 1831: why? This limits the type of address to IPv4.

src/transport/xqc_cid.c
```diff
@@ -248,7 +249,8 @@ xqc_cid_set_get_largest_seq_or_rpt(xqc_cid_set_t *cid_set, uint64_t path_id)
     if (inner_set) {
         return inner_set->largest_scid_seq_num;
     }
-    return XQC_ERROR;
+    /* if the path id is larger than expected, should return 0 here */
```
yangfurong, line 252: What is the reason to change the return value?

src/transport/xqc_cid.c
```diff
@@ -142,6 +142,7 @@ xqc_cid_set_inner_init(xqc_cid_set_inner_t *cid_set_inner)
     xqc_memzero(cid_set_inner, sizeof(xqc_cid_set_inner_t));
     xqc_init_list_head(&cid_set_inner->cid_list);
     xqc_init_list_head(&cid_set_inner->next);
+    cid_set_inner->largest_retire_prior_to = 0;
```
yangfurong, line 145: no need to do this. memzero is already called.

src/transport/xqc_conn.c
```diff
@@ -3790,10 +3790,10 @@ xqc_conn_update_flow_ctl_settings(xqc_connection_t *conn)
 }
 
 xqc_int_t
-xqc_conn_add_path_cid_sets(xqc_connection_t *conn, uint32_t start, uint32_t end)
+xqc_conn_add_path_cid_sets(xqc_connection_t *conn, uint64_t start, uint64_t end)
```
yangfurong, line 3793: why? path id is less than 2^32-1

src/transport/xqc_frame.c
```diff
@@ -2026,6 +2041,53 @@ xqc_process_max_path_id_frame(xqc_connection_t *conn, xqc_packet_in_t *packet_in
...
+    }
+
+    if (conn->local_max_path_id > max_path_id) {
+        xqc_log(conn->log, XQC_LOG_DEBUG,
+                "|received out-dated path blocked frame|local_max_path_id:%ui|max_path_id:%ui|", conn->local_max_path_id, max_path_id);
+        return XQC_OK;
+    }
+
+    uint64_t pre_max_path_id = conn->local_max_path_id;
+    conn->local_max_path_id += (conn->local_max_path_id + 1) / 2;
```
yangfurong, line 2074: We should add some limitations here. For example, it should refuse to increase this value once some thresholds are reached.

src/transport/xqc_multipath.c
```diff
@@ -1487,4 +1491,157 @@ xqc_path_recent_loss_rate(xqc_path_ctx_t *path)
...
+    }
+
+    xqc_log(path->parent_conn->log, XQC_LOG_DEBUG, "|xqc_conn_trigger_cid_rotation_on_path|path_id:%ui|new_dcid:%s|old_dcid:%s|",
+            path_id,
+            xqc_dcid_str(conn->engine,&new_dcid),
+            xqc_scid_str(conn->engine,&path->path_dcid));
+
+    xqc_cid_copy(&path->path_last_dcid, &path->path_dcid);
+    xqc_cid_copy(&path->path_dcid, &new_dcid);
+    new_dcid_inner = xqc_cid_in_cid_set(&conn->dcid_set, &new_dcid, path_id);
```
yangfurong, line 1518: xqc_get_ununsed_cid has already switched the state to CID_USED.

src/transport/xqc_conn.c
Yanmei-Liu: We use u64 for all path ids to avoid casting type everywhere.

yangfurong (approved):
LGTM

PR #468, opened 2024-12-04
fix vulnerablilities

dreamwind1985:
1、NULL pointer dereference
2、Heap OOB read and Stack OOB read when copying a string for logs
3、limit crypto stream frame data buffer and unack packets send buffer


Yanmei-Liu (approved):
Looks good.

PR #472, opened 2025-01-23
 use siphash to replace hash_string

dreamwind1985:
use siphash to replace xqc_hash_string, in order to fix xqc_hash_string vulnerablilities.


Yanmei-Liu (approved):
Looks good!

PR #484, opened 2025-05-06
[~] fix: demo_client is unable to process empty FIN for h3 streams

yangfurong:


yangfurong:
Bugfix for demo_client

PR #503, opened 2025-09-02
[=]: optimize the gf inversion calculation

penglei0:
Optimize the gf inversion calculation:

- 1. do inversion calculation by lookup table
- 2. inversion table is generated by the following python code:

```python
import logging
import galois

def gen_inverse_table(degree: int = 8, irreducible_poly: str = "x^8 + x^4 + x^3 + x^2 + 1"):
    """
        Generate GF(2^degree) inverse table and append as a std::array<uint8_t, N> to output_cpp_file.
    """
    GF = galois.GF(2**degree, irreducible_poly=irreducible_poly)
    size = 2**degree
    inv = [0] * size
    one = GF(1)
    for a in range(1, size):
        inv[a] = int(one / GF(a))  # 1/a in GF
    array_str = ", ".join(str(x) for x in inv)
    c_block = f"static const unsigned char xqc_galois_inv_table[{size}] = {{ {array_str} }};\n\n"
    logging.info("Inverse table generated for GF(2^%d) with irreducible polynomial %s",
                 degree, irreducible_poly)
    logging.info("Inverse table: %s", c_block)


if __name__ == '__main__':
    logging.basicConfig(level=logging.INFO)
    gen_inverse_table()
```


cherylsy (approved):
Looks good.

PR #516, opened 2025-11-16
[Feat]: Support MOQ draft-14 stable

Sy0307:
**Most feat keep stable for MOQ-draft-14. Basic tests have offered and more tests is coming.**

MOQ-Draft-14 link: https://www.ietf.org/archive/id/draft-ietf-moq-transport-14.html

Expected changes or updates compared to master:

- Subscribe series
- Publish series
- Setup series
- Subgroup series

Target: Interop with https://github.com/Sy0307/moxygen successfully.


Sy0307:
Now we support datagram & GOAWAY.

Sy0307:
Update: added server-side MoQ `cancel_write` support.

This commit adds `xqc_moq_track_cancel_write()`, which allows the publisher side to actively close already-open data streams by group filter. The main target scenario is server-side interruption: when the server decides that the previous audio response is no longer useful, it can reset old streams that may still be queued in the MoQ write buffer or QUIC send queue, instead of waiting for TRACK_END/FIN.

Main changes:
- Added public API: `xqc_moq_track_cancel_write()`.
- Added track-level write stream registration and group-based cancel logic.
- Wired media, raw subgroup, and datachannel send paths into the shared cancel path.
- `BEFORE` filters skip future writes for stale groups, avoiding continued transmission of data from an already-cancelled round.
- `EXACT` filters only close currently-open matching streams and do not permanently drop future writes for the same group, avoiding unintended impact on later audio writes within the current group.
- Added e2e test: `test_server_cancel_write_e2e.sh`, which triggers server-side audio `cancel_write` through the demo server.

Usage references:
- API declaration: `include/moq/xqc_moq.h`
- Demo server-side `cancel_write` trigger: `moq/demo/xqc_moq_demo_server.c`
- E2E coverage: `moq/tests/test_server_cancel_write_e2e.sh`

PR #517, opened 2025-11-20
fix bugs when parsing the ack ext frame

dreamwind1985:
When parsing the contents of an ACK ext frame, if a range_count value of 0 is encountered, the packet is considered abnormal and is ignored.


Yanmei-Liu (approved):
Looks good.

Issue #518, opened 2025-11-20
[Bug]: xquic regularly and consistently fails Interop Runner runs


marten-seemann:
### What happened?

The xquic image runs into timeouts and fails the interop runner runs. This happens every time (and triggers an email notification to me). Example run: https://github.com/quic-interop/quic-interop-runner/actions/runs/19495022675

Can you please investigate and fix this?

### Steps To Reproduce

Run the interop runner.

### Relevant log output

```shell

```

marten-seemann:
I just merged the PR. Might be easier to debug if you run it locally though.

Sy0307:
Thanks for merging the PR, but I still do not find the error log on https://interop.seemann.io/. But CI showed that the log files had been uploaded. Can you give some help?

Sy0307:
Have fixed the issue on xquic with neqo in #520 , but xquic still have interoperability issue on kwik & quinn. We will keep researching on the root cause.

Issue #511, opened 2025-11-04
[Bug]: xqc_hq_parse_req|247||hq recv CR LF|


larseggert:
### What happened?

The quic-interop-runner `zerortt` test case with `neqo` as a client and `xquic` at the server times out after several minutes. I tried to debug this, but on the `xquic` side, all I see in the logs is the line
```
xqc_hq_parse_req|247||hq recv CR LF|
```
which repeats thousands of times. On the `neqo` side, this is the last packet received from `xquic`, with nothing afterwards:
```
0.871 DEBUG [Client 904506e79861148e] pn=1 type=Handshake pri-path:c5cee15db60a03f6f41bd99a 0.0.0.0:44601->193.167.100.100:443 Tos(Cs0, NotEct) len 81
  -> RX Crypto { offset: 142, len: 36 }
```
i.e., this is the final server handshake packets that installs the application-space keys. (Full log: [log.txt.gz](https://github.com/user-attachments/files/23342675/log.txt.gz))

### Steps To Reproduce

```shell
uv run python3.13 ./run.py -c neqo -s xquic -t zerortt -d
```
with an installation of the interop runner.

### Relevant log output

```shell
See above.
```

Sy0307:
Sorry for dismissing this issue, I will find the root cause and try to fix it. Thanks!
The similar issue is #518.

PR #520, opened 2025-11-24
[+] Fix: Enhance request parsing to avoid hq parsing error

Sy0307:
Related issue: #518 #511

**Issue:** xqc_hq_parse_req uses sscanf("%s %s") without bounds, so a long URL from neqo overruns the 256‑byte buffer. Since xqc_hq_request_recv_req didn’t null‑terminate the buffer, later strlen() walks into uninitialized memory, causing an ASan heap-buffer-overflow and leaving HQ requests stuck after “hq recv CR LF”.
**Fixes:**
Build a bounded format string (e.g. "%15s %255s") before calling sscanf, so both method and path respect the buffer size.
After reading from the QUIC stream, check whether recv_cnt hits the buffer limit; if not, append '\0'. This guarantees strlen()/sscanf() operate on a valid C string.

When neqo client interop with xquic server, xquic server will trigger core dump & timeout, gdb log like this:
 
```
=================================================================
==8799==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x615000002080 …
    #3 0x100f51c80 in xqc_hq_parse_req xqc_hq_request.c:234
    #4 0x100f52aac in xqc_hq_request_recv_req xqc_hq_request.c:297
    #5 0x100f58600 in xqc_demo_svr_hq_req_read_notify demo_server.c:632
    
```

This PR will solve the issue of parsing error if server run as expected.

### DIFF
--- a/demo/xqc_hq_request.c
+++ b/demo/xqc_hq_request.c
@@ -233,25 +233,23 @@ xqc_hq_parse_req(xqc_hq_request_t *hqr, char *res, size_t sz, uint8_t *fin)
     char method[16] = {0};
     char fmt[32] = {0};
     size_t method_cap = sizeof(method) - 1;
-    size_t res_cap;
-    int ret;
-    size_t request_line_len;
+
 
     if (sz <= 1) {
         PRINT_LOG("|invalid resource buffer size|sz:%zu|", sz);
         return -XQC_EPROTO;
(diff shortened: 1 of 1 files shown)


Yanmei-Liu (approved):
Solve interop test error. Looks good.

Issue #523, opened 2025-12-11
Reporting a potential vulnerability: No response so far


NErinola:
I would like to report a potential vulnerability I identified in the project.

I first reached out via email to xquic@alibaba-inc.com on 30 July 2025, but did not receive any reply. I later submitted the details through GitHub Security on 13 October 2025, but I have not received any response there either.

It would be helpful if someone could check the report submitted via GitHub Security (ID: GHSA-3x9f-rg77-hvch) and let me know whether the vulnerability should be reported through a different channel or if there is an alternative contact or procedure for responsible disclosure.

Thank you!

PR #524, opened 2025-12-12
[+] Fix: Handle illegal STREAM frames in INIT and HSK packets

Sy0307:
Fix the [issue](https://github.com/alibaba/xquic/security/advisories/GHSA-3x9f-rg77-hvch)  mentioned in #523 

The original description is shown below (written by @NErinola) :

> Description
Description of the Issue
The XQUIC implementation currently accepts STREAM frames in both Initial and Handshake packets. It appears to buffer the received application data and sends a response after handshake completion in a 1‑RTT packet. Accepting STREAM frames in Initial or Handshake packets violates the QUIC specification and allows arbitrary application data injection.
Impact
A man‑in‑the‑middle attacker can use Initial packats to inject arbitrary application data into a new QUIC connection before the handshake fully completes. This can lead to:
Injection of arbitrary data that the client/server will process.
Potential confusion of the application protocol, escalating into further vulnerabilities (e.g. resource confusion in HTTP/3).
Mitigation
According to the QUIC specification, STREAM frames must only be processed from 0‑RTT or 1‑RTT packets, never from Initial and Handshake packets.
Tested Implementation
XQUIC v1.8.3 with BabaSSL v8.3-stbale
The official demo server with HTTP3 support

QUIC RFC in [here](https://www.rfc-editor.org/rfc/rfc9000.html#section-12.4-6):

```
0x08-0x0f | STREAM | Section 19.8 | __01 | F
```
So QUIC RFC do not allow stream frame in Handshake/Initial ptks.

### DIFF
--- a/src/transport/xqc_frame.c
+++ b/src/transport/xqc_frame.c
@@ -447,6 +447,16 @@ xqc_process_stream_frame(xqc_connection_t *conn, xqc_packet_in_t *packet_in)
     xqc_stream_t        *stream = NULL;
     xqc_stream_frame_t  *stream_frame;
 
+    if (packet_in->pi_pkt.pkt_type == XQC_PTYPE_INIT
+        || packet_in->pi_pkt.pkt_type == XQC_PTYPE_HSK)
+    {
+        xqc_log(conn->log, XQC_LOG_ERROR,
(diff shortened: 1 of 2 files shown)


Yanmei-Liu (approved):
Looks good.

PR #532, opened 2026-01-20
[=] Optimized test scripts.

p4lik4ri:
Maintain existing support for yum and POSIX sh while adding compatibility with bash and apt-based systems.

Update case_test.sh and xquic_test.sh to run correctly across a wider range of environments.


p4lik4ri:
@Yanmei-Liu can you review the code change please?

PR #535, opened 2026-01-29
[+] Fix:  Support more specific idletimeout calculation

Sy0307:
As RFC 9000 wrote [here](https://www.rfc-editor.org/rfc/rfc9000.html#name-idle-timeout) :

> Each endpoint advertises a max_idle_timeout, but the effective value at an endpoint is computed as the minimum of the two advertised values (or the sole advertised value, if only one endpoint advertises a non-zero value). By announcing a max_idle_timeout, an endpoint commits to initiating an immediate close ([Section 10.2](https://www.rfc-editor.org/rfc/rfc9000.html#immediate-close)) if it abandons the connection prior to the effective value.
> 
> ....
>
> The closing and draining connection states exist to ensure that connections close cleanly and that delayed or reordered packets are properly discarded. These states SHOULD persist for at least three times the current PTO interval as defined in [[QUIC-RECOVERY](https://www.rfc-editor.org/rfc/rfc9000.html#QUIC-RECOVERY)].

### DIFF
--- a/src/transport/xqc_conn.c
+++ b/src/transport/xqc_conn.c
@@ -943,6 +943,9 @@ xqc_conn_create(xqc_engine_t *engine, xqc_cid_t *dcid, xqc_cid_t *scid,
         goto fail;
     }
 
+    // Initial here as idle_timeout dependency
+    xqc_init_list_head(&xc->conn_paths_list);
+
     xqc_conn_init_timer_manager(xc);
 
     if (xc->conn_settings.datagram_redundant_probe) {
@@ -5818,17 +5821,6 @@ xqc_conn_get_local_transport_params(xqc_connection_t *conn, xqc_transport_params
                      conn->original_dcid.cid_buf, conn->original_dcid.cid_len);
         params->original_dest_connection_id_present = 1;
 
-        xqc_gen_reset_token(&conn->original_dcid, 
-                            params->stateless_reset_token,
-                            XQC_STATELESS_RESET_TOKENLEN, 
-                            conn->engine->config->reset_token_key, 
-                            conn->engine->config->reset_token_keylen);
-        params->stateless_reset_token_present = 1;
-
-        xqc_log(conn->log, XQC_LOG_INFO, "|generate sr_token[%s] for cid[%s]",
-                xqc_sr_token_str(conn->engine, params->stateless_reset_token),
-                xqc_scid_str(conn->engine, &conn->original_dcid));
-
     } else {
         params->original_dest_connection_id_present = 0;
     }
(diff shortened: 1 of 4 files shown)


Yanmei-Liu:
Read through this against RFC 9000. A few notes.

The sr_token rebinding (xqc_conn.c around 5832) is a real fix, not just refactoring. Per §10.3 ¶5 and §18.2 the token has to be associated with the CID the server *selected* during the handshake — i.e. `initial_source_connection_id`. Binding it to `original_dcid` never matched what a peer would look up: pre-Retry it sometimes worked by accident, post-Retry it was always broken end to end. Good that this is going in.

The `<= 21` → `< 21` in `xqc_packet_parse_stateless_reset` is also correct. `XQC_STATELESS_RESET_PKT_MIN_LEN` is 21 (xqc_defs.h:44), which matches §10.3 framing (5 bytes + 16-byte token). The old check was rejecting the documented minimum-length valid packet.

On the 3*PTO floor — I disagree with the suggestion to drop it. RFC 9000 §10.1 last paragraph is pretty explicit:

> To avoid excessively small idle timeout periods, endpoints MUST increase the idle timeout period to be at least three times the current Probe Timeout (PTO).

That MUST applies to the idle timeout itself, not to the §10.2 closing/draining timer (separate constant, different purpose). The floor is in the right place.

The send-side restart in `xqc_send_ctl_on_packet_sent` looks right too — `last_ae_send_time <= last_recv_time` correctly encodes the §10.1 ¶2 condition ("no ack-eliciting packet has been sent since the last successful receive"). The 0/0 initial state correctly lets the first ae send rearm the timer.

Things I think should be addressed before this lands:

1. Tests. Four protocol-touching changes, zero coverage in this PR. Per the convention here protocol changes need unit tests. What I'd want to see at minimum:
   - `tests/unittest/xqc_conn_test.c` — idle_timeout calc: `min(local, remote)`; `local==0` falls back to default; `remote==0` falls back to local; 3*PTO floor active vs inactive; ceil rounding (pto=334us should produce 2ms, not 1ms).
   - `tests/unittest/xqc_conn_test.c` — timer restart: first ae send after recv rearms; second ae send without intervening recv must not rearm; pure ACK packet should not trigger the reset block.
   - `tests/unittest/xqc_tp_test.c` — server with `initial_scid` emits token bound to it; client must never emit; post-Retry token derives from `initial_scid`, not `retry_scid` or `original_dcid`.
   - `tests/unittest/xqc_packet_test.c` — `parse_stateless_reset` at buf_size = 20 (reject), 21 (accept — this is the off-by-one regression guard), 64 (accept).

2. `xqc_init_list_head(&conn->conn_paths_list)` at xqc_conn.c:945 is now duplicated — the same init lives in `xqc_conn_init_paths_list` at xqc_multipath.c:539. Pick one source of truth. Probably keep the new early init (timer-manager init depends on it) and drop the later one.

3. The comment `// Initial here as idle_timeout dependency` is hard to parse. Something like `// init before xqc_conn_init_timer_manager — get_idle_timeout → get_max_pto iterates this list` would actually help the next reader.

Out of scope for this PR but related: `idle_timeout * 1000` at the new call site (and the existing one at xqc_engine.c:1317) can overflow uint64 if a peer advertises a near-max varint `max_idle_timeout`. Cleanest fix is clamping at `xqc_decode_max_idle_timeout` so both sites are protected. Worth a follow-up issue.

With tests and the two cleanups I think this is ready to go.

PR #547, opened 2026-03-25
Update packet type for new_packet in FEC scheme

penglei0:
Set packet type to XQC_PTYPE_SHORT_HEADER for 1-rtt data. I believe this patch can fix the failure tests in following log:

```log
check fec recovery function of stream using XOR ...>>>>>>>> pass:0
check fec recovery function of stream using RSC ...>>>>>>>> pass:0
check fec recovery function of stream using PM ...>>>>>>>> pass:0
check fec recovery when send repair packets ahead ...>>>>>>>> pass:0
```
log from https://github.com/alibaba/xquic/actions/runs/23532512561/job/68499532016

The recovered packets will be dropped due to this commit<https://github.com/alibaba/xquic/commit/4764604a0e487eeb49338b4498aecda2194eae84>


cherylsy (approved):
LGTM.

PR #548, opened 2026-04-03
Add peer_tp check before calling transport callback

20083017:
maybe dangling pointer to ‘peer_tp’


Sy0307:
Please do not pull your vibe request.

PR #556, opened 2026-04-26
fix(fec): free repair_key and recv_mask in fec_recv_rpr_syb_list cleanup

mp0rta:
`xqc_create_rpr_symbol` calls `xqc_calloc` 4 times (struct + payload + repair_key + recv_mask), 
https://github.com/alibaba/xquic/blob/64b8df3ac3f64111eb9e00be1a952ba5b07144bb/src/transport/xqc_fec.c#L1066-L1072
but `xqc_fec_ctl_destroy`'s `fec_recv_rpr_syb_list` cleanup loop frees only payload + struct. `repair_key` and `recv_mask` leak whenever the connection closes with any unflushed repair symbol still in the recv list.

## Trigger
Any FEC-negotiated connection that receives a repair frame and then closes — the symbol typically hasn't moved out of the recv list yet, so it leaks.
Caught by ASan in a downstream consumer.


Yanmei-Liu (approved):
Clean, minimal, correct fix. Thanks for catching this with ASan — much appreciated.

PR #558, opened 2026-05-03
fix: breaks multipath and xqc_conn_flag_t bits 32+ on MSVC

### LINKED ISSUE #557: [Bug]: Plain enum constants > INT_MAX rely on a non-portable extension; breaks multipath and `xqc_conn_flag_t` bits 32+ on MSVC
### Summary

Several `enum` types in xquic hold values exceeding `INT_MAX`. Per the C11 standard, this is actually a constraint violation:

- **6.7.2.2 ¶3**: "The expression that defines the value of an enumeration constant shall be an integer constant expression that has a value representable as an `int`."
- **6.4.4.3**: "An identifier declared as an enumeration constant has type `int`."
https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf

GCC and Clang silently relax this constraint as a non-standard extension and widen the underlying type to fit, which is why this hasn't surfaced on Linux. MSVC follows the constraint more strictly:
https://learn.microsoft.com/en-us/cpp/c-language/c-enumeration-declarations?view=msvc-170
>In ANSI C, the expressions that define the value of an enumerator constant always have int type. That means the storage associated with an enumeration variable is the storage required for a single int value. 

 it issues a `C4309` warning and truncates the value to its low 32 bits. Either way, code that depends on an enum constant > `INT_MAX` is non-portable.

### Affected definitions

| File | Symbol | Value | Impact |
|---|---|---|---|
| `xqc_transport_params.h` | `XQC_TRANSPORT_PARAM_INIT_MAX_PATH_ID_V10` | `0x0f739bbc1b666d09` | TP ID encoded as low 32 bits → `multipath_version` negotiation always fails |
| `xqc_transport_params.h` | `XQC_TRANSPORT_PARAM_PMTUD_OPTIONS` | `0x0e08a234ff112300` | Same — PMTUD TP not recognized |
| `xqc_errno.h` | `TRA_MP_PROTOCOL_VIOLATION` | `0x1001d76d3ded42f3` | CONNECTION_CLOSE error code truncated, peers cannot identify it |
| `xqc_conn.h` `xqc_conn_flag_t` | bits 32..42 | `1ULL << 32..42` | `conn->conn_flag |= FLAG` becomes `|= 0`. `MP_WAIT_MP_READY` (38) and `MP_READY_NOTIFY` (39) never set → `ready_to_create_path_notify` never fires → secondary multipath paths cannot be created. Also breaks `PMTUD_PROBING`, `HANDSHAKE_DONE_SENT`, `SERVER_ACCEPT`, `RETRY_SENT` etc. |

### Proposed fix

Move the affected constants out of `enum` into the portable forms — `#define` macros cast to `uint64_t` for the transport-param IDs and the multipath error code, and `typedef uint64_t` + `#define` for `xqc_conn_flag_t`. This is the portable way to hold > `INT_MAX` integer constants in C11. (C23 adds `enum X : uint64_t { ... }` which would be a more direct fix, but it's not yet usable for an xquic baseline.)

A C11 `_Static_assert` is added on `xqc_conn_flag_t` (sizeof + critical bit positions) so future refactors that re-narrow the type fail loudly at compile time.

### Steps To Reproduce

1. Build xquic on Windows with MSVC
2. Run a multipath-enabled client against any standard xquic server with multiple paths

Observed: handshake completes, `xqc_conn_enable_multipath` returns `1|1`, `MP_NEW_CONN_ID`/`ACK_MP` frames flow normally, but `ready_to_create_path_notify` never fires and the WLB scheduler reports `n_paths:1`. Bits 32+ of `conn->conn_flag` are silently dropped at the `|=` assignment because the storage type is `int`-wide.

### Environment

- xquic `main` 64b8df3
- Reproduced with: MSVC 19.x (Visual Studio 2022) on Windows 11

### Relevant log output

```shell

```

mp0rta:
fix #557


mp0rta:
https://github.com/alibaba/xquic/blob/64b8df3ac3f64111eb9e00be1a952ba5b07144bb/src/transport/xqc_transport_params.c#L274
`xqc_put_zero_length_param` takes `xqc_transport_param_id_t` and has the same potential hazard if a 64-bit ID is ever passed to it. No callers do today, but worth considering — possibly as part of a broader refactor of the `xqc_transport_param_id_t` vs `uint64_t` boundary.

mp0rta:
This PR doesn't address it, but I think a fix for xqc_frame_type_t is also necessary.
https://github.com/alibaba/xquic/issues/534
may be `XQC_FRAME_BIT_SID``XQC_FRAME_BIT_REPAIR_SYMBOL`` XQC_FRAME_BIT_NUM` were broken.

yangfurong (approved):
LGTM

Yanmei-Liu (approved):
Excellent analysis and a remarkably clean fix. Thanks for the detailed root-cause writeup with C11 standard references — this is exactly the kind of contribution we love to see.

PR #638, opened 2026-05-12
fix issue 596: use compile-time constants for retry integrity tag key/nonce length

### LINKED ISSUE #596: [Bug]: Retry完整性标签使用strlen获取二进制密钥长度
### What happened?

RFC 9001: Section 5.8 - Retry完整性标签使用AEAD_AES_128_GCM计算
xqc_tls_cal_retry_integrity_tag xqc_tls.c:806-808 使用strlen()获取密钥/nonce长度，脆弱实现
ngtcp2修复指导: 使用ngtcp2_strlen_lit()宏（编译时计算长度），确保密钥/nonce长度始终正确

### Steps To Reproduce

### Relevant log output

```shell

```

dreamwind1985:
Use compile-time constants for retry integrity tag key/nonce length instead of hardcoded values.

Closes #596

### DIFF
--- a/src/tls/xqc_tls.c
+++ b/src/tls/xqc_tls.c
@@ -11,9 +11,9 @@
 #include <openssl/rand.h>
 #include <openssl/hmac.h>
 
-/* RFC 9001 §5.8: use fixed lengths (not strlen) since binary keys may contain '\0' */
-#define XQC_RETRY_INTEGRITY_KEY_LEN     16  /* RFC 9001 §5.8: K = 128 bits */
-#define XQC_RETRY_INTEGRITY_NONCE_LEN   12  /* RFC 9001 §5.8: N = 96 bits  */
+/* RFC 9001 Section 5.8: fixed binary lengths, not strlen(). */
+#define XQC_RETRY_INTEGRITY_KEY_LEN     16  /* K = 128 bits */
+#define XQC_RETRY_INTEGRITY_NONCE_LEN   12  /* N = 96 bits */
 
 
 typedef enum xqc_tls_flag_e {
@@ -801,6 +801,10 @@ xqc_tls_cal_retry_integrity_tag(xqc_log_t *log,
 {
     xqc_int_t ret = XQC_OK;
 
+    if (ver < XQC_IDRAFT_INIT_VER || ver >= XQC_VERSION_MAX) {
+        return -XQC_TLS_INVALID_ARGUMENT;
+    }
+
     xqc_crypto_t *crypto = xqc_crypto_create(XQC_TLS13_AES_128_GCM_SHA256, log);
     if (crypto == NULL) {
         xqc_log(log, XQC_LOG_ERROR, "|create retry crypto error|");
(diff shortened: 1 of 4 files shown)


Yanmei-Liu (commented):
## Review

Core fix is correct and minimal. Two follow-ups recommended below; nits noted but none blocking.

### Why the change is correct

- RFC 9001 §5.8 fixes K=128 bits and N=96 bits as AEAD_AES_128_GCM invariants (RFC 5116 §5.1), not as policy. `strlen()` on a binary key was always wrong in spirit, even when it happened to give the right answer.
- The new RFC 9001 Appendix A.4 ground-truth vector matches verbatim: ODCID `8394c8f03e515708`, SCID `f067a5502a4262b5`, token `746f6b656e` (`"token"`), expected tag `04a265ba2eff4d829058fb3f0f2496ba`. Pseudo-packet construction matches §5.8 Figure 8.
- Zero observable on-wire regression on supported versions. V1 key `be0c…c84e` and nonce `4615…25bb`, plus the draft-29 entries in `src/tls/xqc_tls_common.h:53,71`, contain no `0x00` byte — the old code was correct by accident, the new code is correct by construction.

### Follow-ups (not blocking)

**1. The new test does not actually regress the bug it claims to regress.**

Reverting this patch and re-running `xqc_test_retry_integrity_tag_rfc9001` would still pass: V1 key/nonce contain no `0x00`, so the old `strlen()` returns exactly 16/12 anyway. The test is a useful RFC A.4 conformance check, but as a guard against re-introducing `strlen()` on a binary buffer it has zero mutation power. Two ways to make it actually regression-test the fix:

- A negative test calling `xqc_tls_cal_retry_integrity_tag(..., XQC_IDRAFT_INIT_VER)`. That slot's key/nonce are all-zero (`src/tls/xqc_tls_common.h:55-56, 73-74`), which is the exact case where the old strlen-based code returned 0 and the AEAD silently truncated. Asserting the function produces a deterministic non-empty tag (or a documented error) locks the contract.
- Or a thin wrapper accepting explicit key/nonce lengths, exercised with a synthetic NUL-bearing key against a precomputed AEAD reference output.

**2. Add a `ver` bounds check at function entry.**

`xqc_tls_cal_retry_integrity_tag` (`src/tls/xqc_tls.c:798`) indexes `xqc_crypto_retry_key[ver]` without validating `ver < XQC_VERSION_MAX`. Today the only caller, `xqc_conn_gen_retry_integrity_tag` at `src/transport/xqc_packet_parser.c:914`, passes `conn->version`, which is bound by version negotiation — so it isn't reachable today. But defense-in-depth on a control-plane function is essentially free:

```c
if (ver < 0 || ver >= XQC_VERSION_MAX) {
    xqc_log(log, XQC_LOG_ERROR, "|invalid version|ver:%d|", ver);
    return -XQC_TLS_INVALID_ARGUMENT;
}
```

### Nits

- `tests/unittest/xqc_retry_test.c` ends without a trailing newline (`\ No newline at end of file` in the diff).
- Consider co-locating `XQC_RETRY_INTEGRITY_KEY_LEN` / `_NONCE_LEN` next to the array declarations in `src/tls/xqc_tls_common.h` so a future contributor adding a version slot sees the data and its required dimensions in one place.
- `XQC_RETRY_INTEGRITY_TAG_LEN` (already used at `src/transport/xqc_packet_parser.c:917`) is the AEAD output length and is conceptually distinct from the key length. A one-line comment noting that the three constants are all `16` for AES-128-GCM but bound by different RFC clauses (key — RFC 9001 §5.8 / RFC 5116 §5.1; tag — RFC 5116 §5.1) would help the next reader.

### Out of scope for this PR

- RFC 9369 (QUIC v2) ground truth. The current `xqc_crypto_retry_key[]` has no v2 entry, so v2 retry-tag generation is not reachable through this code path. Worth tracking when/if v2 retry is added.
- Switching storage from `const char * const []` to `const unsigned char [][16]` would let a `_Static_assert(sizeof(arr[0]) == 16)` lock the widths at compile time. Worth doing as a separate refactor.

LGTM after the trailing newline. The `ver` bounds check and a placeholder-version regression test would be welcome here or in a follow-up issue.

dreamwind1985:
https://github.com/alibaba/xquic/pull/755 already merged

PR #640, opened 2026-05-12
fix issue 607: use H3_MISSING_SETTINGS error code when first control stream frame is not SETTINGS

### LINKED ISSUE #607: [Bug]: 控制流第一帧非SETTINGS错误码不正确
### What happened?

RFC 9114: Section 6.2.1 - 必须视为H3_MISSING_SETTINGS
xqc_h3_stream_process_control xqc_h3_stream.c:745-749 使用H3_FRAME_UNEXPECTED而非H3_MISSING_SETTINGS
ngtcp2修复指导: 正确返回H3_MISSING_SETTINGS错误码

### Steps To Reproduce

### Relevant log output

```shell

```

dreamwind1985:
Use H3_MISSING_SETTINGS error code when first frame on control stream is not SETTINGS.

Closes #607


Yanmei-Liu (commented):
## Review

Production change is correct, RFC 9114 §6.2.1 conformant, and minimal. Test verified to catch the regression (built locally, then reverted production hunk and confirmed the new test fails). LGTM in principle. Two follow-ups recommended below before merge or as immediate follow-up.

### Why the change is correct

- RFC 9114 §6.2.1: "If the first frame of the control stream is any other frame type, this MUST be treated as a connection error of type H3_MISSING_SETTINGS." The new conditional at `src/http3/xqc_h3_stream.c:745-746` implements exactly that scope.
- RFC 9114 §8.1 Table 4 fixes `H3_MISSING_SETTINGS = 0x010A`. xquic's enum value at `include/xquic/xqc_errno.h:63` matches.
- Wire-level emission preserved: `XQC_H3_CONN_ERR` stamps `conn->conn_err = 0x10A`, which is emitted verbatim by `xqc_write_conn_close_to_packet` at `src/transport/xqc_conn.c:3272` — no code translation, interop holds.
- Mutation-tested locally: reverting only the production hunk causes the new test to fail at `tests/unittest/xqc_h3_test.c:537` (`conn->conn_err == expected_err`). The test has real regression power.

### Recommended before merge (or immediate follow-up)

**1. Extend coverage to all non-SETTINGS frame types.**
The fixture `xqc_test_h3_missing_settings_one(frame_type, expected_err)` already accepts a frame type parameter and supports DATA. Only the public driver at line 552 is narrow. Add cases for DATA, HEADERS, CANCEL_PUSH, PUSH_PROMISE, MAX_PUSH_ID, and at least one reserved/grease type (e.g. `0x21`) — all of which §6.2.1 covers identically.

**2. Tighten the return-value assertion.**
`tests/unittest/xqc_h3_test.c:536` asserts `processed < 0`. A future refactor returning a different negative errno would still pass while silently losing the H3_MISSING_SETTINGS contract. Suggest `CU_ASSERT(processed == -XQC_H3_MISSING_SETTINGS);`.

**3. Drop or fix the dead branches in the test wrapper.**
`xqc_test_h3_missing_settings_one` accepts a SETTINGS branch (line 508) that, if invoked, would fail the line-536 assertion (valid first SETTINGS returns ≥ 0). Either remove the branch or make the wrapper polymorphic over pass/fail outcomes.

### Out of scope, but worth tracking

- `src/http3/xqc_h3_stream.c:762-764` (second-SETTINGS) and `:771` (H3_SETTINGS_ERROR) still return `-H3_FRAME_UNEXPECTED` / `-H3_SETTINGS_ERROR` directly without going through `XQC_H3_CONN_ERR`. Because those negative values (-261, -265) don't pass the macro's `<= -800` guard, even if invoked the wire stamp would be skipped. Per RFC 9114 §7.2.4 these paths are also `MUST close with` cases. Recommend a follow-up PR mirroring this PR's internal-errno + `XQC_H3_CONN_ERR` pattern across the rest of `xqc_h3_stream_process_control`.

- Pre-existing latent issue: at line 745 the check uses `pctx->frame.type` which is `XQC_H3_FRM_UNKNOWN` (UINT64_MAX) immediately after reset. A peer sending only the leading byte of a multi-byte varint type can trigger MISSING_SETTINGS on an incomplete frame. Standard types all fit in one byte so practical reach is narrow, but worth a defensive guard (`pctx->state == XQC_H3_FRM_STATE_END`).

### Nits

- The `extern ssize_t xqc_h3_stream_process_control(...)` redeclaration at `tests/unittest/xqc_h3_test.c:470` belongs in `src/http3/xqc_h3_stream.h` (gated by `XQC_ENABLE_TESTING` if you want to keep it out of the public ABI).
- Unreachable `ret = XQC_ERROR;` at line 518 (`CU_FAIL_FATAL` longjmps).

LGTM after items 1-3.

dreamwind1985:
[Build / Build on Ubuntu (ubuntu-latest) (pull_request)]失败原因是pr771引入，与本次改动无关

Yanmei-Liu (approved):
Looks good.

PR #643, opened 2026-05-12
fix issue 609: reject SETTINGS frame received on http3 request stream

### LINKED ISSUE #609: [Bug]: http3 非控制流上的SETTINGS帧未拒绝
### What happened?

RFC 9114: Section 7.2.4 - 收到时应视为H3_FRAME_UNEXPECTED
xqc_h3_stream_process_request xqc_h3_stream.c:858-1053 请求流上收到SETTINGS被静默忽略
ngtcp2修复指导: 非控制流上收到SETTINGS帧时以H3_FRAME_UNEXPECTED关闭连接

### Steps To Reproduce

### Relevant log output

```shell

```

dreamwind1985:
Reject SETTINGS frame received on http3 request stream with H3_FRAME_UNEXPECTED error and add unit tests.

Closes #609


Yanmei-Liu (commented):
## Review

Blocking: this PR breaks `make run_tests` on the current branch. Two of the three new tests fail because they assert behaviour from fixes that haven't landed yet.

### Build/run result

Built locally with BoringSSL, ran the unit-test suite:

```
xqc_test_h3_critical_stream_close            ...FAILED  (xqc_h3_test.c:529, :531)
xqc_test_h3_data_headers_on_control_rejected ...FAILED  (xqc_h3_test.c:681, :682 x2)
xqc_test_h3_settings_on_request_rejected     ...passed
```

`git log src/http3/xqc_h3_stream.c` on this branch shows only the two #609 commits. There is no production code adding the `XQC_H3_STREAM_TYPE_CONTROL` guard / `H3_CLOSED_CRITICAL_STREAM` raise in `xqc_h3_stream_close_notify` (issue #606), and no `case XQC_H3_FRM_DATA` / `HEADERS` arms in `xqc_h3_stream_process_control` (issue #612). The new tests for those issues will remain RED until those fixes land.

### Required before merge

**1. Resolve the test/code mismatch.** Either:
- drop `xqc_test_h3_critical_stream_close` and `xqc_test_h3_data_headers_on_control_rejected` from this PR until #606 / #612 production fixes are merged, or
- include those production fixes in this PR and rescope the title/description accordingly.

**2. Remove the internal-process leftover at `tests/unittest/xqc_h3_test.c:739`:**
> `* Per user decision TestStrategy=B, this is the simplified 2-case variant:`

Replace with a normal rationale comment (e.g. "Only positive SETTINGS + DATA regression case; §7.2.4 only mandates SETTINGS rejection in this PR's scope").

### Issues with the production fix itself (#609)

**A. Internal errno mismatch.** `src/http3/xqc_h3_stream.c:1042-1044` reuses `XQC_H3_CONTROL_ERROR` (`include/xquic/xqc_errno.h:244`, defined as *"control stream error, such as setting not send first or send twice"*). The new arm runs on a **request** stream; the comment now lies. Wire emission is unaffected (the first macro arg, `H3_FRAME_UNEXPECTED`, is what reaches the wire), but the internal errno will mislead future debuggers. The sibling PR #638 minted a dedicated `XQC_H3_MISSING_SETTINGS` for an analogous case — recommend the same pattern here (e.g. `XQC_H3_REQUEST_FRAME_UNEXPECTED`) rather than stretching `CONTROL_ERROR`.

**B. Production fix is narrower than RFC §7.2 mandates.** Per RFC 9114:
- §7.2.3 CANCEL_PUSH on non-control stream MUST be H3_FRAME_UNEXPECTED
- §7.2.6 GOAWAY on non-control stream MUST be H3_FRAME_UNEXPECTED
- §7.2.7 MAX_PUSH_ID on non-control stream MUST be H3_FRAME_UNEXPECTED

The current `default:` arm at `src/http3/xqc_h3_stream.c:1046-1049` silently ignores all of these. RFC 9114 §9 ("ignore unknown frame types") covers genuinely unknown types, not well-known frames received on the wrong stream. Suggest extending the `case` ladder to explicitly reject the three above with `H3_FRAME_UNEXPECTED`, keeping `default:` for the §9 ignore-unknown path. Mutation-test pattern would mirror the existing SETTINGS test.

### Coverage suggestion

`xqc_test_h3_settings_on_request_rejected` only covers the SETTINGS positive case and a DATA regression guard. Once the fix is broadened per (B), extend the test to cover CANCEL_PUSH, GOAWAY, MAX_PUSH_ID positive cases plus a HEADERS regression guard. The fixture `xqc_test_h3_settings_on_request_one(frame_type, expected_err)` is already parameterised — only the public driver needs new entries.

### Nits (not blocking)

- Three `extern` declarations of exported-but-not-headered functions in `tests/unittest/xqc_h3_test.c`. Consolidate into `src/http3/xqc_h3_stream.h` under `#ifdef XQC_ENABLE_TESTING`. Same hack pattern is replicating across recent PRs.
- Unreachable `ret = XQC_ERROR;` after `CU_FAIL_FATAL` (~line 30, ~658, ~797).
- Pre-existing engine leak in test scaffolding (`test_engine_connect` never destroys the returned engine). Out of scope here, but worth a separate cleanup PR.
- `include/xquic/xqc_configure.h` shows as modified in the working tree but isn't in the diff — local CMake artefact. Add to `.gitignore` if not already.

Once the failing tests are resolved (per option 1 or 2 above), the SETTINGS-rejection logic itself is RFC-conformant and the unit test for it is mutation-verified to catch the regression. LGTM after the test/code mismatch is resolved and the comment hygiene fixed.

dreamwind1985:
issue  609在https://github.com/alibaba/xquic/pull/796这个PR中解决

PR #644, opened 2026-05-12
fix issue 612: reject DATA and HEADERS frames received on http3 control stream

### LINKED ISSUE #612: [Bug]:  http3 DATA帧在控制流上未拒绝
### What happened?

RFC 9114: Section 7.2.1 - DATA帧不得在控制流上发送
xqc_h3_stream_process_control xqc_h3_stream.c:720-813 DATA/HEADERS在控制流上被静默忽略
ngtcp2修复指导: 控制流上收到DATA帧时以H3_FRAME_UNEXPECTED关闭连接

### Steps To Reproduce

### Relevant log output

```shell

```

dreamwind1985:
Reject DATA and HEADERS frames received on http3 control stream with H3_FRAME_UNEXPECTED error and add unit tests.

Closes #612


Yanmei-Liu:
A few things on this before it can land.

The biggest one: this PR also adds `xqc_test_h3_critical_stream_close` for #606, but the #606 fix itself isn't on this branch and isn't on main. I grepped both — `H3_CLOSED_CRITICAL_STREAM` is raised nowhere outside `xqc_errno.h`. Case 1 of that test asserts `conn->conn_err == H3_CLOSED_CRITICAL_STREAM`, which will fail as soon as the unittest runs after merge. Also probably explains why the PR is `CONFLICTING` — both branches inject into the same lines of `xqc_h3_test.c`, `xqc_h3_test.h`, and `main.c`. Cleanest is to drop the #606 test bits from here and let them ship with #606. Otherwise this needs to be rebased on top of #606 with the dependency called out in the description.

The actual #612 fix only covers DATA and HEADERS, but RFC 9114 §7.2.5 imposes the same MUST on PUSH_PROMISE:

> a PUSH_PROMISE frame received on a control stream MUST cause the receiver to treat this as a connection error of type H3_FRAME_UNEXPECTED.

Today PUSH_PROMISE on the control stream gets parsed all the way through (`xqc_h3_frame.c:244-247` runs `xqc_h3_frm_parse_push_promise`, `fin` is set, state → END) and then drops into the `default:` arm of the dispatch switch in `xqc_h3_stream.c:808-812` which just logs `ignore unknown frame`. Same shape of bug as #612, same RFC subsection family, not fixed. Either extend the new guard:

```c
if (pctx->state >= XQC_H3_FRM_STATE_PAYLOAD
    && (pctx->frame.type == XQC_H3_FRM_DATA
        || pctx->frame.type == XQC_H3_FRM_HEADERS
        || pctx->frame.type == XQC_H3_FRM_PUSH_PROMISE))
```

or — preferably — fold the rejection into the existing dispatch switch so the allowlist of control-stream-legal frames lives in one place rather than being split between an inline `if` and the `default:` case. The test would want renaming too (`xqc_test_h3_request_frames_on_control_rejected` or similar) plus a PUSH_PROMISE positive case.

Smaller things worth doing in the same pass:

- The first commit's comment explained why the check has to run before the `state != END` gate (parser doesn't consume payload for DATA/HEADERS so state never advances to END). The terser version in the second commit drops that, which makes future regressions easier.
- Positive-case assertions only check `processed < 0`. Worth pinning the exact code (`processed == -XQC_H3_CONTROL_ERROR`) and that pctx was actually reset (`pctx.state == XQC_H3_FRM_STATE_TYPE`).
- Coverage in the current 4 cases is thin: zero-length DATA (the guard catches it today but nothing pins that), partial-feed across multiple `process_control` calls, a concatenated `GOAWAY || DATA` buffer (verify the loop still rejects after a legal frame and that GOAWAY's side effects landed first), and DATA without pre-setting `XQC_H3_CONN_FLAG_SETTINGS_RECVED` (verifies the new guard takes precedence over the #607 SETTINGS-first guard).

Separate from this PR, the inverse direction — SETTINGS/GOAWAY/MAX_PUSH_ID/CANCEL_PUSH arriving on a request stream — is governed by §7.2.3 / §7.2.4 / §7.2.6 / §7.2.7 with the same MUST. `xqc_h3_stream_process_request` likely has the symmetric parser-stall pathology that produced #612 here. Worth a separate issue.

Yanmei-Liu (approved):
The current version looks good.

PR #757, opened 2026-05-21
Add a length check for cid when parsing transport parameters

dreamwind1985:
fix TP CID decode heap overflow when CID's param_len > XQC_MAX_CID_LEN


Yanmei-Liu (approved):
LGTM.

PR #758, opened 2026-05-21
fix issue 559: use min(local, remote) max_idle_timeout per RFC 9000 10.1

### LINKED ISSUE #559: [Bug]: 空闲超时时间取较小值，实现未对齐RFC协议描述【严重】
### What happened?

rfc9000中
10.1. 空闲超时（Idle Timeout）
如果任一终端在其传输参数（第18.2节）中指定max_idle_timeout，则当连接保**持空闲的时间超过两端通告的max_idle_timeout值的最小值时，连接将被静默关闭并丢弃其状态**。xquic实现没有这部分处理
xqc_conn_get_idle_timeout() 第6396行 return conn->local_settings.max_idle_timeout == 0 ? XQC_CONN_DEFAULT_IDLE_TIMEOUT : conn->local_settings.max_idle_timeout，完全忽略远端max_idle_timeout

开源实现ngtcp2中有对应协议处理
ngtcp2_tstamp ngtcp2_conn_get_idle_expiry(ngtcp2_conn *conn) {
  ngtcp2_duration trpto;
  ngtcp2_duration idle_timeout;

  /* TODO Remote max_idle_timeout becomes effective after handshake
     completion. */

  if (!conn_is_tls_handshake_completed(conn) ||
      conn->remote.transport_params->max_idle_timeout == 0 ||
      (conn->local.transport_params.max_idle_timeout &&
       conn->local.transport_params.max_idle_timeout <
         conn->remote.transport_params->max_idle_timeout)) {
    idle_timeout = conn->local.transport_params.max_idle_timeout;
  } else {
    idle_timeout = conn->remote.transport_params->max_idle_timeout;
  }

  if (idle_timeout == 0) {
    return UINT64_MAX;
  }

  trpto = 3 * conn_compute_pto(conn, conn_is_tls_handshake_completed(conn)
                                       ? &conn->pktns
                                       : conn->hs_pktns);

  idle_timeout = ngtcp2_max_uint64(idle_timeout, trpto);

  if (conn->idle_ts >= UINT64_MAX - idle_timeout) {
    return UINT64_MAX;
  }

  return conn->idle_ts + idle_timeout;
}

### Steps To Reproduce

### Relevant log output

```shell

```

Yanmei-Liu:
RFC 9000 §10.1 requires the effective idle timeout to be the minimum of the `max_idle_timeout` values advertised by both endpoints, where 0 means "no limit". The current implementation in `xqc_conn_get_idle_timeout()` (`src/transport/xqc_conn.c:6466`) only reads `local_settings.max_idle_timeout` and ignores `remote_settings.max_idle_timeout` entirely. When the local side advertises a longer timeout than the peer, the connection can survive past the peer's silently-discarded state, leading to traffic black-holing.

## What this PR does

- Read `remote_settings.max_idle_timeout` after handshake completion
- Apply `min(local, remote)` with `0` treated as "no limit" (per RFC 9000 §18.2)
- Fall back to `XQC_CONN_DEFAULT_IDLE_TIMEOUT` when both sides advertise 0, so the idle timer is never disabled entirely
- Preserve the server pre-handshake `init_idle_time_out` path as a DDoS mitigation, unrelated to this fix

## Behaviour matrix

| local | remote | post-handshake result |
|---|---|---|
| 30000 | 5000 | 5000 (min) |
| 0 | 30000 | 30000 (peer's value) |
| 30000 | 0 | 30000 (peer disabled) |
| 0 | 0 | XQC_CONN_DEFAULT_IDLE_TIMEOUT (safety fallback) |

## Tests

Added `tests/unittest/xqc_conn_test.c::xqc_test_conn_idle_timeout` with 10 cases covering the negotiation matrix, pre-handshake behaviour for both client and server roles, equal values and the no-overflow case. All 10 new asserts pass; existing test suite remains green.

## Future work (not in this PR)

- RFC 9000 §10.1 also requires `idle_timeout >= 3 * PTO`. xquic does not currently track this floor anywhere; adding it touches the PTO computation path and is left for a separate change.
- An explicit `xqc_timer_set(XQC_TIMER_CONN_IDLE, ...)` at handshake completion would close a small theoretical window where a server's `init_idle_time_out` timer still runs against the larger negotiated value. The next received packet already resets the timer in `xqc_engine.c:1317-1318`, so no real bug observed.

Fixes: #559


yangfurong (approved):
LGTM

PR #761, opened 2026-05-21
fix issue 606: detect peer close of HTTP/3 critical streams per RFC 9114 6.2.1 / RFC 9204 4.2

### LINKED ISSUE #606: [Bug]: 控制流关闭不检测为H3_CLOSED_CRITICAL_STREAM [严重]
### What happened?

RFC 9114 Section 6.2.1 - 控制流关闭必须视为H3_CLOSED_CRITICAL_STREAM
xqc_h3_stream_process_data xqc_h3_stream.c:1965-2018 无控制流关闭检测，H3_CLOSED_CRITICAL_STREAM(0x104)从未使用
ngtcp2修复指导: nghttp3在控制流关闭时立即以H3_CLOSED_CRITICAL_STREAM错误关闭连接

### Steps To Reproduce

### Relevant log output

```shell

```

Yanmei-Liu:
RFC 9114 §6.2.1 and RFC 9204 §4.2 require an endpoint to treat the closure of any HTTP/3 critical stream (the control stream, the QPACK encoder stream, or the QPACK decoder stream) as a connection error of type `H3_CLOSED_CRITICAL_STREAM` (0x104). xquic currently ignores this entirely — `xqc_h3_stream_close_notify()` does not check the stream type at all, and the 0x104 error code defined in `xqc_errno.h` was never raised.

## What this PR does

- Add a small `xqc_h3_stream_is_critical()` helper so all three stream types are checked in one place; future critical streams (e.g. WebTransport) only need to extend this one function.
- In `xqc_h3_stream_close_notify()`, when a critical stream's transport close fires while the connection is still healthy, log the violation and raise `H3_CLOSED_CRITICAL_STREAM` via `XQC_H3_CONN_ERR`.
- Skip the report when the connection is already shutting down (`XQC_CONN_FLAG_CLOSING_NOTIFY`) or has reached `XQC_CONN_STATE_CLOSING`, so a local close or a transport-level error doesn't get clobbered by a spurious critical-stream error during teardown.

## Why no `ACTIVELY_CLOSED` guard

A previous attempt added an `XQC_HTTP3_STREAM_FLAG_ACTIVELY_CLOSED` check to distinguish local-initiated closes. It turns out that flag is only set by `xqc_h3_stream_close()`, which is never called for control or QPACK streams (those are owned by `xqc_h3_conn_t` and torn down through `xqc_h3_conn_destroy` instead). On the critical-stream path the flag is always false, so the branch is unreachable. The `CLOSING_NOTIFY` and `conn_state` guards already cover every legitimate teardown scenario, so this PR keeps the logic flat — one `if`, three guards, no dead branches.

## Tests

`tests/unittest/xqc_h3_test.c::xqc_test_h3_critical_stream_close` is a table-driven test with 9 cases:

- 3 positive cases: peer-close of CONTROL / QPACK_ENCODER / QPACK_DECODER on a healthy connection raises 0x104.
- 3 negative cases: REQUEST / PUSH / UNKNOWN streams do not.
- 2 teardown-guard cases: CLOSING_NOTIFY flag and `conn_state == CLOSING` both suppress the report.
- 1 first-write-wins case: a pre-existing `conn_err` is preserved.

All 9 pass; the existing test suite remains green. The fix was also mutation-tested: collapsing the helper to only check CONTROL fails the two QPACK cases, and removing the `conn_state` guard fails the `state_closing` case — confirming each branch is actually exercised.

Fixes: #606


yangfurong (approved):
LGTM

PR #762, opened 2026-05-21
fix issue 727: enforce RFC 9002 §7.2 cap on NewReno initial window

### LINKED ISSUE #727: [Bug]: NewReno初始拥塞窗口缺少RFC上限约束
### What happened?

RFC 9002 Section 7.2: “limiting the window to the larger of 14,720 bytes or twice the maximum datagram size”

XQC_kInitialWindow 定义为 10 * XQC_kMaxDatagramSize，但没有实现RFC要求的 min(10*max_datagram_size, max(2*max_datagram_size, 14720)) 上限约束。当MSS=1436时恰好合规（14360 < 14720），但如果通过路径MTU发现后MSS增大（如MSS=9000的jumbo frame），则IW = 90000字节，远超RFC允许的14720字节上限

在大MTU网络中，初始窗口可能过大，导致突发流量过多
xqc_new_reno.c:18
ngtcp2的IW = 10 * 1200 = 12000，同样缺少上限约束逻辑，但1200字节的MSS使得IW始终小于14720

### Steps To Reproduce

### Relevant log output

```shell

```

Yanmei-Liu:
RFC 9002 §7.2 RECOMMENDS limiting the initial congestion window to `min(10 * max_datagram_size, max(2 * max_datagram_size, 14720))`. The previous `XQC_kInitialWindow` macro only computed the `10 * MSS` term; with the current `XQC_kMaxDatagramSize = XQC_MSS = 1436` the 14720-byte ceiling is not binding (10 * 1436 = 14360), so the macro happened to return a compliant value. A build configured with a larger maximum datagram size (for example jumbo frames at MSS=9000) would have computed `90000` bytes, well above the `18000`-byte cap the RFC permits in that case.

Express the full RFC formula at the point of definition using the shared `xqc_min` / `xqc_max` helpers from `src/common/xqc_config.h` and drop the file-local `xqc_max` redefinition. With the default MSS the resulting value is unchanged, so behaviour on stock builds is preserved.

The sibling CCAs in `src/congestion_control/` (CUBIC, BBR, BBR2, Copa) default to `32 * MSS = 45952` bytes — over the RFC ceiling even at the stock MSS — but that is out of scope for issue #727 and tracked separately.

> Side note for the issue reporter: ngtcp2 does implement the §7.2 formula — see [`lib/ngtcp2_cc.c` `ngtcp2_cc_compute_initcwnd`](https://github.com/ngtcp2/ngtcp2/blob/main/lib/ngtcp2_cc.c#L38). Worth correcting the original issue body.

### Tests

New CUnit function `xqc_test_reno_init_cwnd`:

- live initial window equals the formula evaluated against the compile-time `XQC_MSS` (regression guard for this issue)
- a formula helper covers canonical MSS values 1200 / 1436 / 1500 / 9000 so a build configured with a different maximum datagram size cannot silently drift past the RFC bound — the 9000 case in particular exercises the 18000-byte cap the original macro failed to honour

All five assertions pass against `./tests/run_tests`.

Fixes: #727


yangfurong (approved):
LGTM

PR #764, opened 2026-05-22
fix issue 599: use peer-advertised max_ack_delay in PTO estimate

### LINKED ISSUE #599: [Bug]:  PTO计算使用错误的max_ack_delay【严重】
### What happened?

RFC 9002: Section 6.2.1 - PTO应使用对端报告的max_ack_delay
xqc_send_ctl_calc_pto xqc_send_ctl.h:182 使用local_settings.max_ack_delay而非remote_settings
ngtcp2修复指导: 使用conn->remote.transport_params->max_ack_delay，Initial/Handshake空间设为0

static inline xqc_usec_t
xqc_send_ctl_calc_pto(xqc_send_ctl_t *send_ctl)
{
return send_ctl->ctl_srtt + xqc_max(4 * send_ctl->ctl_rttvar, XQC_kGranularity * 1000)
+ send_ctl->ctl_conn->local_settings.max_ack_delay * 1000;
}
使用了 local_settings.max_ack_delay（本端的 max_ack_delay），而非 remote_settings.max_ack_delay（对端的 max_ack_delay）。RFC 明确要求使用对端的 max_ack_delay，因为 PTO 是用来等待对端发送 ACK 的超时。

影响：PTO 计算使用了错误的 max_ack_delay 值，可能导致 PTO 过长或过短。

ngtcp2修复指导:
函数: ngtcp2_conn_compute_pto() (ngtcp2_conn.c) — 被调用点在 rtb_detect_lost_pkt() (ngtcp2_rtb.c:1055)
处理流程: PTO计算公式（RFC 9002第6.2节）：PTO = smoothed_rtt + max(4 * rttvar, kGranularity) + max_ack_delay * (2 ^ pto_count)。其中max_ack_delay来自远端传输参数 conn->remote.transport_params->max_ack_delay。Initial/Handshake空间max_ack_delay设为0。pto_count每次收到ACK时重置为0(ngtcp2_rtb.c:744)。持久拥塞周期计算中也使用相同的远端max_ack_delay(ngtcp2_rtb.c:1079-1081)

### Steps To Reproduce

### Relevant log output

```shell

```

Yanmei-Liu:
Fixes: #599

### What

`xqc_send_ctl_calc_pto` was computing the PTO term with
`local_settings.max_ack_delay`, but per RFC 9002 §6.2.1 the formula
must use the peer-reported value. PTO is the budget *we* wait for
the *peer* to ACK, so the peer's advertised delay is what bounds it.

In stock interop the bug stays hidden because both endpoints often
advertise the 25 ms default, so the formula collapses to the right
answer. With a peer that advertises something different the local
side declares the peer's packets lost too early or too late.

### Fix

```c
return send_ctl->ctl_srtt
     + xqc_max(4 * send_ctl->ctl_rttvar, XQC_kGranularity * 1000)
     + send_ctl->ctl_conn->remote_settings.max_ack_delay * 1000;
```

Three sibling sites in `xqc_send_ctl.c` already use
`remote_settings.max_ack_delay`:

- line 1159 — ACK-delay clamp on RTT sample (§5.3)
- line 1439 — persistent congestion duration (§7.6)
- line 1685 — pns-aware PTO arming (§6.2.1)

The helper at `xqc_send_ctl.h:182` was the lone outlier. The log
line at `xqc_send_ctl.c:1651` already prints
`c->remote_settings.max_ack_delay`, which is good internal
evidence that `remote` is the intended source of truth.

### Why not a bigger refactor

The Initial/Handshake "`max_ack_delay` term must be 0" rule is
the responsibility of `xqc_send_ctl_get_pto_time_and_space`, which
already gates on `pns == XQC_PNS_APP_DATA`. Plumbing a `pns`
parameter through `calc_pto` would touch 5+ files and the only
caller (`xqc_conn_get_max_pto`) deliberately wants a 1-RTT-space
worst case across paths for idle / close timers. Adding pns
granularity at that aggregation layer is the wrong abstraction.

I added a comment at the helper documenting that contract so the
next reader does not get surprised.

### Pre-handshake

`remote_settings.max_ack_delay` is initialized to
`XQC_DEFAULT_MAX_ACK_DELAY = 25` by `xqc_conn_set_default_settings`
before any path / send_ctl exists, so the formula never reads
garbage. Pre-handshake the result is conservative-larger, which
is the safe direction for idle-timer floors.

### Tests

Two CUnit tests in `tests/unittest/xqc_send_ctl_test.c`:

- `xqc_test_pto_uses_remote_max_ack_delay` — sets `local=25`,
  `remote=100`, pins `srtt=1000us`, `rttvar=0` so the floor term
  collapses to `XQC_kGranularity * 1000`. Asserts the formula
  consumes `remote * 1000`, not `local * 1000`.
- `xqc_test_pto_remote_default_when_unset` — defensive guard that
  the formula yields a sane positive value at default settings.

I ran the negative-control sanity check: temporarily reverted the
swap, rebuilt, re-ran. The first test failed at both its
assertions; restoring the patch made both tests pass. The test
actually exercises the regression.

### Out of scope

While auditing I noticed two adjacent items, leaving them for
separate issues:

- `XQC_kGranularity` is 2 ms in `xqc_send_ctl.h:26`; RFC 9002 §A.2
  fixes it at 1 ms. Likely a deliberate tuning choice but
  undocumented.
- `xqc_decode_max_ack_delay` (`xqc_transport_params.c:621`) has
  no `< 2^14` clamp; per RFC 9000 §18.2 a peer advertising a
  larger value MUST be rejected with `TRANSPORT_PARAMETER_ERROR`.
  Same exposure exists at the three sibling sites that already
  use `remote_settings`; this PR doesn't widen it.

### DIFF
--- a/src/transport/xqc_send_ctl.h
+++ b/src/transport/xqc_send_ctl.h
@@ -175,11 +175,21 @@ typedef struct xqc_send_ctl_s {
 } xqc_send_ctl_t;
 
 
+/*
+ * Per-path connection-level PTO estimate used by callers that do not need
+ * pns granularity (e.g. xqc_conn_get_max_pto). Per RFC 9002 6.2.1, PTO must
+ * be computed using the peer-reported max_ack_delay, hence remote_settings.
+ *
+ * The pns-specific Initial/Handshake rule (max_ack_delay term must be 0
+ * before handshake confirmation) is enforced by
+ * xqc_send_ctl_get_pto_time_and_space; callers needing that granularity
+ * must use it instead. See xqc_send_ctl.c for the canonical formula.
+ */
 static inline xqc_usec_t
 xqc_send_ctl_calc_pto(xqc_send_ctl_t *send_ctl)
 {
     return send_ctl->ctl_srtt + xqc_max(4 * send_ctl->ctl_rttvar, XQC_kGranularity * 1000)
-        + send_ctl->ctl_conn->local_settings.max_ack_delay * 1000;
+        + send_ctl->ctl_conn->remote_settings.max_ack_delay * 1000;
 }
 
 
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -113,6 +113,7 @@ if(HAVE_CUNIT)
         ${UNIT_TEST_DIR}/xqc_datagram_test.c
         ${UNIT_TEST_DIR}/xqc_h3_ext_test.c
         ${UNIT_TEST_DIR}/xqc_ack_with_timestamp_test.c
+        ${UNIT_TEST_DIR}/xqc_send_ctl_test.c
     )
 
     if(XQC_ENABLE_FEC)
--- a/tests/unittest/main.c
+++ b/tests/unittest/main.c
@@ -41,6 +41,7 @@
 #include "xqc_fec_scheme_test.h"
 #include "xqc_fec_test.h"
 #include "xqc_ack_with_timestamp_test.h"
+#include "xqc_send_ctl_test.h"
 
 static int xqc_init_suite(void) { return 0; }
 static int xqc_clean_suite(void) { return 0; }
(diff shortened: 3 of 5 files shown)


dreamwind1985 (approved):
LGTM. RFC 9002 §6.2.1 明确要求 PTO 使用对端的 max_ack_delay，修复正确。测试用正反验证覆盖了回归路径。代码库中其他 3 处已正确使用 remote_settings，本处是唯一遗漏。CI 全过。

PR #768, opened 2026-05-22
fix issue 724: cap ack_delay by default max_ack_delay before handshake confirmation

### LINKED ISSUE #724: [Bug]:  握手确认前不裁剪ack_delay，可能导致srtt被高估
### What happened?

RFC 9002 Section 5.3: “Before handshake confirmation, the peer is expected to use the default max_ack_delay value, so ack_delay should be capped by the default value”

xquic 在 xqc_conn_is_handshake_confirmed 为false时不裁剪ack_delay，直接使用ACK帧中报告的ack_delay值。RFC 9002 Section 5.3的伪代码中，ack_delay始终被 max(max_ack_delay, ack_delay) 裁剪，不区分握手是否确认。在握手确认前，对端可能使用默认的max_ack_delay，而xquic使用ACK帧中报告的值（可能不准确），导致RTT估计偏差

在握手阶段，如果对端报告的ack_delay异常大（例如由于实现bug），xquic不会裁剪它，可能导致srtt被低估（因为adjusted_rtt = latest_rtt - 过大的ack_delay）

ngtcp2 在 ngtcp2_conn_update_rtt 中始终裁剪ack_delay，不区分握手状态

### Steps To Reproduce

### Relevant log output

```shell

```

Yanmei-Liu:
RFC 9002 §5.3 says an endpoint SHOULD ignore the peer's `max_ack_delay` until the handshake is confirmed, and MUST use the lesser of the acknowledgement delay and the peer's `max_ack_delay` after that. xquic only applied the clamp in `xqc_send_ctl_update_rtt()` once the handshake was confirmed — before that the value the peer put in the ACK frame went straight into `adjusted_rtt = latest_rtt - ack_delay`, unbounded.

Most of the time this is harmless because `remote_settings.max_ack_delay` is initialised to the spec default (25ms, see `xqc_conn_set_default_settings`), so the post-confirmation path would clamp to the same value anyway. The window where it matters is 0-RTT resumption: `xqc_conn_set_early_remote_transport_params` overwrites `remote_settings.max_ack_delay` with the cached value from the previous session, which can be much larger. If an early ACK arrives with a big `ack_delay`, `adjusted_rtt` ends up far below `latest_rtt`, srtt is underestimated, PTO shrinks, and we trigger spurious loss detection.

## What this PR does

- In the handshake-not-confirmed branch, cap `ack_delay` by `XQC_DEFAULT_MAX_ACK_DELAY` (25ms, RFC 9000 §18.2) instead of leaving it unbounded. Using the constant directly — rather than relying on `remote_settings.max_ack_delay` already holding the default — keeps the behaviour correct even when the cached transport parameters from a previous 0-RTT session have a non-default value.
- The post-confirmation branch is unchanged.
- The plausibility check below (`adjusted_rtt + 1000 >= minrtt + ack_delay`) is left alone; its purpose and the 1ms fudge are unrelated to this fix.

## Tests

`tests/unittest/xqc_send_ctl_test.c::xqc_test_send_ctl_update_rtt_ack_delay_cap` is a table-driven test with 8 cases:

- Handshake not confirmed, large `ack_delay` (100ms) → clamped to 25ms.
- Handshake confirmed, large `ack_delay` → clamped to negotiated `remote_settings.max_ack_delay`.
- Small `ack_delay` (below either cap) → unchanged.
- First RTT sample shortcut (sets srtt directly, no cap).
- `ack_delay == 0` corner.
- Plausibility guard rejects subtraction when it would drop below `min_rtt`.
- Post-confirmation with `remote_settings.max_ack_delay == 0` → cap is 0, subtraction skipped.

All 8 pass; existing test suite is green. The fix was mutation-tested: removing the new `else` branch fails case 1, and dropping the `* 1000` unit conversion fails it too — confirming both the branch and the unit math are actually exercised.

Fixes: #724

### DIFF
--- a/src/transport/xqc_send_ctl.c
+++ b/src/transport/xqc_send_ctl.c
@@ -21,6 +21,7 @@
 #include "src/transport/xqc_utils.h"
 #include "src/transport/xqc_datagram.h"
 #include "src/transport/xqc_reinjection.h"
+#include "src/transport/xqc_transport_params.h"
 
 int 
 xqc_send_ctl_may_remove_unacked_dgram(xqc_connection_t *conn, xqc_packet_out_t *po)
@@ -1155,8 +1156,19 @@ xqc_send_ctl_update_rtt(xqc_send_ctl_t *send_ctl, xqc_usec_t *latest_rtt, xqc_us
     } else {
         send_ctl->ctl_minrtt = xqc_min(*latest_rtt, send_ctl->ctl_minrtt);
 
+        /*
+         * RFC 9002 5.3: ack_delay MUST be clamped by max_ack_delay before
+         * being subtracted from latest_rtt. Until the handshake is
+         * confirmed the peer is required to use the default max_ack_delay
+         * (RFC 9000 18.2), so use that constant; after confirmation the
+         * negotiated value from the peer's transport parameters applies.
+         */
         if (xqc_conn_is_handshake_confirmed(send_ctl->ctl_conn)) {
-            ack_delay = xqc_min(ack_delay, send_ctl->ctl_conn->remote_settings.max_ack_delay * 1000);
+            ack_delay = xqc_min(ack_delay,
+                                send_ctl->ctl_conn->remote_settings.max_ack_delay * 1000);
+
+        } else {
+            ack_delay = xqc_min(ack_delay, XQC_DEFAULT_MAX_ACK_DELAY * 1000);
         }
 
         /* Adjust for ack delay if it's plausible. */
--- a/tests/unittest/main.c
+++ b/tests/unittest/main.c
@@ -127,7 +127,9 @@ main()
                         xqc_test_pto_uses_remote_max_ack_delay)
         || !CU_add_test(pSuite, "xqc_test_pto_remote_default_when_unset",
                         xqc_test_pto_remote_default_when_unset)
-        /* ADD TESTS HERE */) 
+        || !CU_add_test(pSuite, "xqc_test_send_ctl_update_rtt_ack_delay_cap",
+                        xqc_test_send_ctl_update_rtt_ack_delay_cap)
+        /* ADD TESTS HERE */)
     {
         CU_cleanup_registry();
         return (int)CU_get_error();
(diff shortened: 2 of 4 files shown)


dreamwind1985 (approved):
LGTM. RFC 9002 §5.3 要求握手确认前使用默认 max_ack_delay (25ms) 裁剪 ack_delay，修复正确。8 个表驱动测试用例覆盖了各种边界场景。注意：本 PR 与 #764 都新建了 xqc_send_ctl_test.c，合并时需 resolve 冲突。CI 全过。

yangfurong (approved):
LGTM

Yanmei-Liu:
Rebased on latest `main` (`origin/main` = `8ababa6`) to clear the conflict.

Conflicts were all in the test scaffolding because #764 (`fix issue 599: use peer-advertised max_ack_delay in PTO estimate`) had already landed on main and created the same `tests/unittest/xqc_send_ctl_test.{c,h}` files this PR was adding. Resolution was to merge both files' contents and keep all three test functions:

- `tests/unittest/xqc_send_ctl_test.h` — both `xqc_test_pto_*` declarations (from #764) alongside `xqc_test_send_ctl_update_rtt_ack_delay_cap` (this PR); kept #764's `XQC_SEND_CTL_TEST_H` include-guard pattern since it matches the convention used by the other `_test.h` files
- `tests/unittest/xqc_send_ctl_test.c` — union of both include sets, then #764's two PTO functions followed by this PR's `xqc_rtt_case_t` / helper / `xqc_test_send_ctl_update_rtt_ack_delay_cap`. The auto-merge had fused the closing braces of `xqc_test_pto_remote_default_when_unset` and `xqc_test_send_ctl_update_rtt_ack_delay_cap` into one because both end with `xqc_engine_destroy(conn->engine)`; split them out.
- `tests/unittest/main.c` — kept all three `CU_add_test` registrations on the same chain

`src/transport/xqc_send_ctl.c` (the actual `xqc_send_ctl_update_rtt` change) auto-merged cleanly.

Re-ran `./tests/run_tests` after the rebase (BoringSSL, macOS): the two #764 tests and the new #768 test all pass.

`git push --force-with-lease`, new tip is `f04d829`.

PR #766, opened 2026-05-22
fix issue 681: reset all 0-RTT streams when early data is rejected

Yanmei-Liu:
RFC 9001 §4.6.2 requires that when 0-RTT is rejected the client MUST reset the state of all streams that were sent in 0-RTT. `xqc_conn_early_data_reject` iterates `conn_all_streams` and resets each stream carrying `XQC_STREAM_FLAG_HAS_0RTT`, but when it encountered a stream already in `RESET_SENT` / `RESET_RECVD` it returned `XQC_OK` from the middle of the loop. Any 0-RTT streams later in the list were then left with their pre-reject `send_offset` / `unacked_pkt` and their buffered writes were not re-sent in 1-RTT, which violates the MUST and can leave data stuck on streams the server has no record of. The caller in `xqc_conn_handshake_complete` discards the return value, so there was no compensating retry path either.

Continue the iteration instead of returning, so every 0-RTT stream is either re-initialised for 1-RTT retransmission or, if it is already reset, has its buffered 0-RTT writes discarded.

### Tests

New `xqc_test_conn_early_data_reject` builds four streams on a client connection — stream 2 forced into `RESET_SENT` and stream 3 left live with non-zero `send_offset` — and asserts that after the call:

- stream 1 (live 0-RTT) is back at READY/RECV with zeroed offsets
- stream 2 (already reset) keeps its terminal state but has its buffered writes dropped
- stream 3 (the regression guard) is reset just like stream 1 — with the pre-fix code it would still hold the original `send_offset`
- stream 4 (no 0-RTT flag) is not touched

All assertions pass against `./tests/run_tests`.

Fixes: #681

PR #769, opened 2026-05-22
fix issue 767: reset connection-level fc_data_sent on 0-RTT reject

### LINKED ISSUE #767: [Bug]: connection-level fc_data_sent not reset on 0-RTT reject double-counts replayed bytes
### What happened?

`xqc_conn_early_data_reject` (`src/transport/xqc_conn.c:4138`) leaves the connection-level send counter `conn->conn_flow_ctl.fc_data_sent` untouched when 0-RTT is rejected. The same buffered application data is then replayed over 1-RTT via `xqc_stream_write_buffed_data_to_packets`, which traverses `xqc_write_stream_frame_to_packet` and unconditionally adds the replayed bytes to `fc_data_sent` again. The connection-level send accounting therefore double-counts every byte that was originally written over 0-RTT.

#### Code path

`fc_data_sent` is the client-side counter that tracks how much data the local endpoint has sent towards the peer's advertised `MAX_DATA`:

```c
/* src/transport/xqc_conn.h:255-257 */
xqc_conn_flow_ctl_t {
    ...
    uint64_t                fc_max_data_can_send;
    uint64_t                fc_data_sent;
};
```

It is incremented inside `xqc_write_stream_frame_to_packet` regardless of packet type:

```c
/* src/transport/xqc_packet_out.c:1247 */
stream->stream_send_offset += *send_data_written;
stream->stream_conn->conn_flow_ctl.fc_data_sent += *send_data_written;
```

`xqc_conn_early_data_reject` runs three things in order:

1. `xqc_send_queue_drop_0rtt_packets(conn);` — releases queued 0-RTT `xqc_packet_out_t`s but does not decrement `fc_data_sent` (`src/transport/xqc_send_queue.c:434-509`).
2. `xqc_conn_resend_0rtt_datagram(conn);`
3. The per-stream loop, which (after PR #766) iterates every `XQC_STREAM_FLAG_HAS_0RTT` stream:
   - resets `stream->stream_send_offset = 0`;
   - calls `xqc_stream_write_buffed_data_to_packets(stream)` to re-emit the buffered writes in 1-RTT.

The replay in step 3 calls `xqc_write_stream_frame_to_packet`, so `fc_data_sent` is incremented a second time for the exact same application bytes.

#### RFC reference

RFC 9001 §4.6.2 ([Accepting and Rejecting 0-RTT](https://www.rfc-editor.org/rfc/rfc9001.html#section-4.6.2)):

> If 0-RTT data is rejected, all connection characteristics that the client assumed might be incorrect. … The client therefore MUST reset the state of all streams, including application state bound to those streams.

The MUST is stated in terms of "the state of all streams". Per-stream `stream_send_offset` is reset accordingly. The connection-level send aggregate `fc_data_sent` is, by construction, the sum of every stream's `stream_send_offset` accumulated over the lifetime of the connection. After the per-stream offsets are reset to zero, leaving the connection-level aggregate at its pre-reject value is internally inconsistent with the per-stream state and breaks the "treat 0-RTT as not sent" intent of §4.6.2 even though the section does not name `fc_data_sent` explicitly.

Note: this is **not** a strict §4.1 MUST violation. RFC 9000 §4.1 forbids sending more than the advertised limit; the bug here is the opposite direction — the local counter is inflated, so the client believes it has consumed more of the peer's `MAX_DATA` than it has and may stop sending early. Severity is "correctness / unnecessary throttling", not "wire-protocol violation".

#### Observable effect

Take a 0-RTT-eligible client whose remembered `initial_max_data` was N bytes. The client buffers and sends N/2 bytes over 0-RTT (`fc_data_sent` becomes N/2). The server rejects 0-RTT and announces a fresh `initial_max_data` of N bytes via 1-RTT transport parameters; xquic updates `fc_max_data_can_send = N` at `src/transport/xqc_conn.c:4294`. The early-data-reject loop then replays the N/2 bytes in 1-RTT, taking `fc_data_sent` to N. The very next stream write hits the `fc_data_sent + pkt_out_size > fc_max_data_can_send` check at `src/transport/xqc_stream.c:354` and emits a `DATA_BLOCKED` frame, even though the peer has only actually accepted N/2 bytes on the connection so far. Throughput on the affected connection is artificially capped at half of the negotiated window until the peer issues `MAX_DATA` increments.

#### Comparison with ngtcp2

ngtcp2's analogous handler `ngtcp2_conn_tls_early_data_rejected` (`lib/ngtcp2_conn.c`) explicitly does the connection-level reset alongside the per-stream destroy:

```c
/* via conn_discard_early_data_state */
conn->tx.offset = 0;
conn->tx.last_blocked_offset = UINT64_MAX;
conn->rx.unsent_max_offset = conn->rx.max_offset = local.initial_max_data;
```

xquic's stream-side mirror was added in PR #766; the connection-side mirror is what this issue tracks.

#### Suggested fix

In the client branch of `xqc_conn_early_data_reject`, after `xqc_send_queue_drop_0rtt_packets` and before the per-stream replay loop, reset `conn->conn_flow_ctl.fc_data_sent = 0`. Replaying the buffered writes in 1-RTT will then re-establish the counter at the correct, single-counted value. A regression test should assert that `fc_data_sent` after the call equals the sum of the post-reject `stream_send_offset` values rather than that sum plus the original 0-RTT volume.

The receive-side counters (`fc_data_recvd`, `fc_max_data`) are not affected — the client only sent on 0-RTT, it did not receive on 0-RTT — but a follow-up audit of every flow-control / congestion-control field touched by 0-RTT-only sends would be worth doing in the same change.

### Steps To Reproduce

1. Open a client connection that uses cached transport parameters with a small `initial_max_data` (e.g. 64 KB).
2. Write the full window of application data over 0-RTT before the handshake completes.
3. Configure the server to reject 0-RTT (e.g. force fresh ticket, mismatched ALPN, or `early_data` extension absent).
4. After handshake completion, attempt to write more application data on the same streams.
5. Observe that `xqc_conn_should_send_data_blocked_to_peer`-style logic is triggered immediately because `fc_data_sent` is at the pre-reject + replay total; verify by logging `conn->conn_flow_ctl.fc_data_sent` and comparing with the sum of `stream_send_offset` across `conn_all_streams`.

### Relevant log output

```shell

```

Yanmei-Liu:
RFC 9001 §4.6.2 requires the client to reset the state of all streams when 0-RTT is rejected. `xqc_conn_early_data_reject` resets each 0-RTT stream's `stream_send_offset` to zero, but the connection-level send aggregate `conn->conn_flow_ctl.fc_data_sent` — which is the sum of every stream's send progress — was left at its pre-reject value.

`xqc_send_queue_drop_0rtt_packets` releases the queued 0-RTT `packet_out` objects but does not decrement `fc_data_sent`, and the buffered application data is then replayed in 1-RTT through `xqc_stream_write_buffed_data_to_packets` → `xqc_write_stream_frame_to_packet`, which adds the same bytes to `fc_data_sent` a second time (`src/transport/xqc_packet_out.c:1247`). The result is that the rejected 0-RTT bytes are charged twice against the peer's advertised `MAX_DATA`. The client may then issue a `DATA_BLOCKED` frame and stop sending well before the real connection limit is reached.

Zero `fc_data_sent` in the client branch after the queued 0-RTT packets have been released and before the per-stream replay loop. The replay path will rebuild the counter to its correct, single-counted value. The reset relies on the fact that `XQC_CONN_FLAG_CAN_SEND_1RTT` is set and `xqc_conn_early_data_reject` is reached in the same synchronous engine cycle (`xqc_conn_check_handshake_complete` calls `xqc_conn_check_tx_key` immediately followed by `xqc_conn_handshake_complete`), so no app-driven 1-RTT write can interleave between the two and be accidentally zeroed.

`xqc_conn_flow_ctl_t`'s other fields were audited and do not need parallel resets: receive-side counters are driven by app reads and must not roll back; `fc_max_data_can_send` and the stream-limit counterparts are reset from fresh transport parameters at `xqc_conn.c:4305`; path-level inflight is cleared by `xqc_send_queue_drop_0rtt_packets` via `xqc_send_ctl_decrease_inflight`; multi-path send-byte counters are observability stats.

Branched from `main`. PR #766 fixes a sibling iteration bug in the same function but on different lines; the two patches do not conflict semantically and a textual rebase will be trivial for whichever lands second.

### Tests

New `xqc_test_conn_early_data_reject_flow_ctl` covers:

- client connection with a HAS_0RTT stream, `fc_data_sent` primed to 4096 — must be back to zero after the call (the regression guard)
- client connection with no HAS_0RTT streams, `fc_data_sent = 8192` — must still be zero (the reset is unconditional in the client branch)
- `conn_type` overridden to `XQC_CONN_TYPE_SERVER`, `fc_data_sent = 16384` — must remain unchanged because the server branch returns before reaching the reset line

All three cases pass against `./tests/run_tests`.

Fixes: #767


dreamwind1985 (approved):
LGTM. RFC 9001 §4.6.2 要求重置所有 0-RTT 流状态，fc_data_sent 作为 stream_send_offset 的连接级聚合必须同步清零，否则 replay 双倍计费导致提前 DATA_BLOCKED。修复精准，一行业务改动。测试覆盖 client-with-0RTT / client-no-0RTT / server-branch 三种场景。与 PR #766 语义独立、行位置不冲突。CI 全过。

yangfurong (approved):
LGTM

PR #770, opened 2026-05-22
fix issue 582: continue processing coalesced packets after a tolerant error

### LINKED ISSUE #582: [Bug]: 合并包解析错误时丢弃剩余字节【严重】
### What happened?

RFC 9000  Section 12.2 - 应丢弃该包但继续处理后续包
xquic 1.9.1 xqc_conn.c:5023-5028 xqc_conn_process_packet goto end跳过所有剩余包
ngtcp2修复指导: 每个包独立解析，错误不传播到同数据报中其他包

ngtcp2修复指导:
函数: ngtcp2_conn_read_pkt() (ngtcp2_conn.c)
处理流程: 当合并包中的某个包解密或处理失败时，ngtcp2不会中止整个UDP数据报的处理，而是跳过失败的包，继续尝试处理剩余的包。每个包独立解析，错误不传播到同数据报中其他包

### Steps To Reproduce

### Relevant log output

```shell

```

Yanmei-Liu:
Fixes: #582

### What

`xqc_conn_process_packet` aborts the entire UDP datagram with
`goto end` as soon as a packet returns one of the tolerant errors
(`XQC_EILLPKT`, `XQC_EDECRYPT`, `XQC_EWAITING`, `XQC_EIGNORE_PKT`,
`XQC_EVERSION`). RFC 9000 §12.2 is explicit:

> if decryption fails (because the keys are not available or for
> any other reason), the receiver MAY either discard or buffer the
> packet for later processing and **MUST attempt to process the
> remaining packets**.

The most concrete impact is during handshake. A client that
coalesces Initial+Handshake into one datagram forces the server
to drop the Handshake packet whenever the Initial path returns
`EWAITING` (e.g. while Handshake keys are still being installed),
costing one full RTT to retransmit.

### Fix

```c
} else if (xqc_conn_tolerant_error(ret)) {
    xqc_log(c->log, XQC_LOG_INFO,
            "|tolerant err, skip packet and continue|%d|", ret);
    pos = packet_in->last > pos ? packet_in->last : end;
    ret = XQC_OK;
    xqc_log_event(c->log, TRA_PACKET_RECEIVED, packet_in);
    continue;
}
```

The ternary guarantees forward progress: if the parser advanced
`packet_in->last` past the failing packet (the long-header
post-Length case), `pos` jumps to the next packet's start; if
not, `pos = end` exits the loop cleanly. The qlog
`TRA_PACKET_RECEIVED` emission mirrors the success path so a
discarded coalesced packet still shows up in event logs.

The non-tolerant branch (`return ret`) is unchanged — those
errors are connection-fatal (`PROTOCOL_VIOLATION` etc.) and should
keep terminating the loop.

### Tests

Two CUnit tests in `tests/unittest/xqc_packet_test.c`:

- `xqc_test_coalesced_continue_after_tolerant_error` — builds a
  two-packet coalesced datagram. Packet 1 is a synthetic Initial
  with a valid header and parseable Length but garbage AEAD body,
  so it fails with `XQC_EDECRYPT` after the parser has set
  `packet_in->last` to the next packet's start. Packet 2 is a
  single byte that returns `XQC_EIGNORE_PKT`. The test asserts
  the function returns OK and the `rcv_pkt_stats.conn_rcvd_pkts`
  counter increments by exactly 2 — proving both packets were
  visited.
- `xqc_test_coalesced_zero_progress_terminates` — feeds a
  one-byte garbage datagram and asserts the function returns OK
  and exits the loop without spinning, exercising the fallback
  branch where `packet_in->last` did not advance.

Reverting the patch flips the first test to FAIL (counter only
increments by 1), confirming it actually exercises the
regression.

### Out of scope

- The non-tolerant branch is connection-fatal by design and
  remains so.
- A bigger refactor that splits `xqc_packet_process_single`'s
  return values into per-packet vs. connection-level taxonomies
  would prevent this whole class of regression. Worth a separate
  RFC.
- macOS-specific pre-existing flake at
  `xqc_test_packet_encrypt_hp_sample_boundary` (engine recreate
  fails after first test). The new tests are registered early in
  the suite to ensure they run on Darwin even if the flake fires
  later. A separate issue is warranted.

### DIFF
--- a/scripts/case_test.sh
+++ b/scripts/case_test.sh
@@ -4947,4 +4947,56 @@ else
     case_print_result "ack_timestamp_frame_case_6" "fail"
 fi
 
+
+# issue #582 regression: a tolerant per-packet error inside a coalesced
+# UDP datagram must let the loop continue to the remaining packets.
+# Pre-fix code emitted an "|ignore err|" INFO log and `goto end` after
+# the first tolerant error, silently dropping every subsequent packet
+# in the same datagram. The fix replaces that with
+# "|tolerant err, skip packet and continue|" and a continue, plus
+# de-duplicates the qlog packet_received emission on the EDECRYPT
+# path (which is already covered by packet_dropped in xqc_packet.c).
+# Assertions:
+#   1. normal HTTP/3 transfer completes,
+#   2. no errors on either side,
+#   3. the legacy "|ignore err|" string -- which only appears in the
+#      buggy code path -- is absent in both clog and slog,
+#   4. when the new "|tolerant err, skip packet and continue|" line
+#      shows up (it does during the regular handshake when keys for
+#      a coalesced Handshake packet are not yet installed) it is
+#      never emitted with ret==-XQC_EDECRYPT alongside a duplicate
+#      qlog packet_received for the same packet number.
+killall test_server 2> /dev/null
+clear_log
+echo -e "coalesced-tolerant continue (issue #582) ...\c"
+${SERVER_BIN} -l d -e -x 53 > /dev/null &
+sleep 1
+${CLIENT_BIN} -s 1024000 -l d -t 2 -E -x 53 > stdlog
+transfer_ok=`grep ">>>>>>>> pass" stdlog`
+errlog=`grep_err_log`
+legacy_log=`grep "|ignore err|" clog slog 2>/dev/null | wc -l | tr -d ' '`
+# If qlog events are written to the log file (depends on test build
+# config), enforce per-log dedup of pkt_num between received/dropped.
+collisions=0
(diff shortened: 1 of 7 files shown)


dreamwind1985 (approved):
LGTM. RFC 9000 §12.2 明确要求 tolerant error 后必须继续处理剩余合并包。修复用 continue + 前进指针替代 goto end，带安全回退防无限循环。测试精准验证了双包前进和零进度终止两个关键路径。CI 全过。

yangfurong (changes requested):
不对，本身就已经是一个完整的QUIC packet而不是UDP packet，pos update 到last才是正确的方案。

PR #778, opened 2026-05-25
fix issue 751: include 32-byte per-field overhead in receive-side header size check

### LINKED ISSUE #751: [Bug]: 接收端Header Size检查计算不准确
### What happened?

RFC 9114  Section 4.2.2:  
“The size of a field list is calculated based on the uncompressed size of fields, including the length of the name and value in bytes plus an overhead of 32 bytes for each field.” 

部分合规，但有bug
xquic在接收端检查header大小时(xqc_h3_request_on_recv_header)使用的是 headers->total_len（仅包含name+value长度），没有加上每个字段32字节的开销。而发送端(xqc_h3_stream.c:174-182)正确使用了 xqc_h3_uncompressed_fields_size 函数（包含32字节开销）。接收端检查失败时仅返回 -XQC_H3_INVALID_HEADER，没有发送431状态码。

xqc_h3_request.c:814-822 使用 headers->total_len 而非 headers->total_len + headers->count * 32 与 max_field_section_size 比较; xqc_h3_stream.c:174-182 发送端正确使用了 xqc_h3_uncompressed_fields_size

nghttp3_conn.c 中 conn_read_request 和 conn_read_response 通过 max_field_section_size 设置检查，计算方式正确包含32字节开销

修复建议：
在接收端使用 xqc_h3_uncompressed_fields_size 函数计算header大小

### Steps To Reproduce

### Relevant log output

```shell

```

Yanmei-Liu:
Fixes: #751

## Background

RFC 9114 Section 4.2.2 defines the size of a field list as the sum of
the uncompressed name and value lengths plus a 32-byte overhead per
field. The send path already accounts for the overhead via
`xqc_h3_uncompressed_fields_size()`; the receive path was comparing
only `headers->total_len` against `SETTINGS_MAX_FIELD_SECTION_SIZE`,
so an oversize field list with many small fields could slip past the
limit by up to `count * 32` bytes.

## Change

* Hoist `xqc_h3_uncompressed_fields_size()` to `xqc_h3_stream.h` so the
  helper is shared by send and receive sides instead of being
  reimplemented (or, in this case, forgotten) at the receive site.
* Use the helper in `xqc_h3_request_on_recv_header()` and log the
  computed `fields_size`.
* `header_recvd` is intentionally left as raw name+value bytes; it
  pairs with `compressed_header_recvd` as a wire-bytes telemetry
  counter, where adding the synthetic 32B per field would distort the
  reported amount.

## On the reporter's 431 note

The reporter also suggested the receive path should emit HTTP 431.
RFC 9114 Section 4.2.2 marks 431 as MAY rather than MUST, and 431 is
an HTTP application-layer response that the upper layer is free to
synthesize on top of the `-XQC_H3_INVALID_HEADER` returned by the
library; keeping that decision out of the transport layer aligns with
nghttp3 / ngtcp2 / quiche. No change there.

## Tests

* `xqc_test_h3_uncompressed_fields_size` — formula direct test
  (empty, single-field, many-field, regression boundary).
* `xqc_test_h3_recv_header_field_section_size` — drives
  `xqc_h3_request_on_recv_header` across the boundary; a pre-fix
  binary accepts `count=1, total_len=80` against a limit of 100, the
  fixed binary rejects it.


yangfurong (approved):
OK

Yanmei-Liu (commented):
Overall the fix looks good — the receive path was missing the 32-byte-per-field overhead that the send path already accounts for. Moving the helper into the shared header is clean, and the test coverage hits the right boundary cases.

One thing to fix before merging: the multiplication in `xqc_h3_uncompressed_fields_size()` runs in `size_t` arithmetic, which is 32-bit on ARM32 targets (Android). If `count` is large enough, `count * 32` wraps before being widened to `uint64_t`. A cast fixes it:

```c
return (uint64_t)headers->total_len + (uint64_t)headers->count * 32;
```

Would also be nice to add a test case with many small fields (e.g. count=50, total_len=50, limit=100) since that's closer to the actual attack pattern from #751 — many tiny headers whose individual sizes look fine but whose aggregate 32B overhead blows the limit.

Everything else checks out: formula matches the RFC, `>` is correct for "exceeds", telemetry counter rightly stays as raw bytes, trailers go through the same check. Nice fix.

PR #777, opened 2026-05-25
Fix issue 776: exclude handshake CIDs from active_connection_id_limit per RFC 9000 §5.1.1

dreamwind1985:
`xqc_cid_set_insert_cid()` incorrectly counted handshake-negotiated CIDs toward `active_connection_id_limit`, violating RFC 9000 §5.1.1 . This caused xquic client to reject valid NEW_CONNECTION_ID frames from nginx (which sends 7 NCIDs with limit=8), resulting in immediate connection close and 0% interop pass rate.
 
Closes #776

PR #780, opened 2026-05-26
patch for #777: harden original CID accounting per review feedback

Yanmei-Liu:
## Summary

Fixup for #777 — addresses review feedback on the handshake CID exclusion implementation:

- **Idempotency guard**: `xqc_cid_set_mark_original` now checks `!inner->is_original` before incrementing counter, preventing silent inflation if called twice on the same CID
- **Sender-side supply alignment**: `xqc_conn_try_add_new_conn_id` now subtracts `original_cid_cnt` in all three `while` conditions, so the endpoint correctly issues up to `active_connection_id_limit` non-handshake CIDs
- **Delete path coverage**: `xqc_cid_set_delete_cid` now decrements `original_cid_cnt` when removing an active original CID
- **File hygiene**: trailing newlines added to `xqc_cid_test.c` and `xqc_cid_test.h`

## Test plan

- [ ] Existing `xqc_test_cid_handshake_exclusion` passes (no behavioral regression — guard only activates on double-call which doesn't occur in current tests)
- [ ] `xqc_test_cid_active_limit` passes (sender-side change is consistent with receiver-side fix)
- [ ] Full test suite: `sh scripts/xquic_test.sh`

Fixes review comments on #777.


Yanmei-Liu:
Close. Fix at original PR.

PR #781, opened 2026-05-26
[~] fix issue 763: honor xqc_cc_params_t.init_cwnd override in NewReno

### LINKED ISSUE #763: [Bug]: NewReno silently ignores xqc_cc_params_t.init_cwnd override
### What happened?

`xqc_reno_init` in `src/congestion_control/xqc_new_reno.c` does not read `xqc_cc_params_t` at all. As a result, callers that set `cc_params.customize_on = 1, cc_params.init_cwnd = N` to override the initial congestion window are silently ignored when the chosen congestion controller is NewReno.

```c
size_t
xqc_reno_size()
{
    return sizeof(xqc_new_reno_t);
}

static void
xqc_reno_init(void *cong_ctl, xqc_send_ctl_t *ctl_ctx,
    xqc_cc_params_t cc_params)
{
    xqc_new_reno_t *reno = (xqc_new_reno_t *)(cong_ctl);
    /* cc_params is never read */
    reno->reno_congestion_window = XQC_kInitialWindow;
    ...
}
```

For comparison, the sibling implementations honour the override:
- CUBIC: `src/congestion_control/xqc_cubic.c:128-137` clamps `cc_params.init_cwnd` into `[XQC_CUBIC_MIN_WIN, XQC_CUBIC_MAX_INIT_WIN]` when `customize_on` is set.
- BBR: `src/congestion_control/xqc_bbr.c:285-294` follows the same pattern.
- Copa: `src/congestion_control/xqc_copa.c:108` follows the same pattern.

A caller writing `cc_params = {.customize_on = 1, .init_cwnd = 20}` will get the configured value on every CCA except NewReno, where it is silently dropped.

Found while reviewing #727. Filing as a separate issue because the fix is a behaviour change (existing callers may be relying on the silent-ignore today) and should be considered on its own merits.

### Steps To Reproduce

1. Open a connection with `xqc_conn_settings_t.cong_ctrl_callback = xqc_reno_cb` and `cc_params = {.customize_on = 1, .init_cwnd = 20}`.
2. Observe `reno_congestion_window` after init equals `XQC_kInitialWindow` (the macro value) rather than `20 * XQC_kMaxDatagramSize`.

### Relevant log output

```shell

```

Yanmei-Liu:
Fixes: #763

## What

`xqc_reno_init` previously ignored the `cc_params` argument, so callers
that set `customize_on = 1` with a tuned `init_cwnd` had the override
silently dropped on NewReno. Cubic, BBR and Copa all honor the same
field, so the behaviour was algorithm-dependent with no log line to tell
the operator.

| CCA | `customize_on=1, init_cwnd=20` | Actual IW |
|---|---|---|
| NewReno (before) | ignored | `XQC_kInitialWindow` |
| Cubic | accepted | `20 * MSS` |
| BBR | accepted | `20 * MSS` |
| Copa | accepted | `20 * MSS` |

## Why this fix

- The `customize_on` contract is exposed in `xqc_cc_params_t` and applies
  across CCAs; one CCA quietly violating it is a contract bug, not an
  intentional design choice.
- Aligning NewReno with the sibling CCAs keeps the numeric `init_cwnd`
  range consistent: the same `[4, 100]` packet range is valid everywhere
  so operators can switch CCAs without re-tuning the input.
- Out-of-range values fall back to the RFC 9002 default rather than being
  silently clamped, so misconfiguration surfaces instead of being
  rewritten. The RFC 9002 Section 7.2 cap on the default itself is still
  enforced by the existing \`XQC_kInitialWindow\` formula.

## What is out of scope

- \`min_cwnd\` override is also dropped on NewReno today, but honoring it
  would require adding a per-instance field to \`xqc_new_reno_t\` and
  threading it through \`xqc_reno_on_lost\` / \`xqc_reno_reset_cwnd\`.
  That is an ABI-visible change and is intentionally tracked separately
  so this PR stays focused on what #763 reports.
- The \`uint32_t init_cwnd * MSS\` multiplication can overflow for
  pathological inputs, but Cubic and BBR have the identical pattern.
  Treating it here only would diverge from the sibling implementations.

## Tests

\`xqc_test_reno_init_cwnd_override\` covers:

- zero-initialised \`cc_params\` -> default IW
- \`customize_on=0\` with non-zero \`init_cwnd\` -> default IW (field ignored)
- \`customize_on=1, init_cwnd=20\` -> \`20 * MSS\`
- exact lower bound (4) and upper bound (100) accepted
- \`3\` / \`101\` / \`0\` -> fallback to default

Pins the \`[4, 100]\` packet contract against future drift and guards
against silent regression of #763.


yangfurong (approved):
OK

PR #782, opened 2026-05-26
[+] fix issue #744: emit H3_MESSAGE_ERROR for malformed HTTP/3 messages per RFC 9114 4.1.2

### LINKED ISSUE #744: [Bug]:  缺少H3_MESSAGE_ERROR错误码定义
### What happened?

RFC9114  Section 4.1.2：
 “Malformed requests or responses that are detected MUST be treated as a stream error of type H3_MESSAGE_ERROR.”

xquic的错误码枚举(xqc_errno.h)中定义了 H3_REQUEST_REJECTED(0x10B)、H3_REQUEST_CANCELLED(0x10C)、H3_REQUEST_INCOMPLETE(0x10D)，但缺少 H3_MESSAGE_ERROR(0x10E) 的定义。当检测到格式错误的请求/响应时，xquic无法发送正确的H3_MESSAGE_ERROR错误码，而是使用 H3_GENERAL_PROTOCOL_ERROR 或内部错误码替代。

xqc_errno.h:46-60 错误码枚举中缺少 H3_MESSAGE_ERROR = 0x10E; xqc_h3_stream.c:1509-1513 xqc_h3_stream_process_in() 对 -XQC_H3_INVALID_HEADER 使用 H3_GENERAL_PROTOCOL_ERROR 而非 H3_MESSAGE_ERROR

nghttp3_http.c:531-563 nghttp3_http_on_header() – 返回 NGHTTP3_ERR_MALFORMED_HTTP_HEADER，最终映射为 NGHTTP3_H3_MESSAGE_ERROR

修复建议：
在 xqc_errno.h 中添加 H3_MESSAGE_ERROR = 0x10E，并在格式错误场景中使用此错误码

### Steps To Reproduce

### Relevant log output

```shell

```

Yanmei-Liu:
## Summary

- Add \`H3_MESSAGE_ERROR = 0x10E\` to \`xqc_h3_err_code_t\` at its
  IANA-assigned slot (RFC 9114 §8.1 Table 2). The enum previously
  skipped from \`H3_REQUEST_INCOMPLETE\` to \`H3_CONNECT_ERROR\`,
  so the value was unreachable from xquic code paths.
- Route the QPACK-decode malformed-header path in
  \`xqc_h3_stream_process_in\` to the new \`H3_MESSAGE_ERROR\` per
  RFC 9114 §4.1.2 (MUST). Previously it fell through to
  \`H3_GENERAL_PROTOCOL_ERROR\`, which §A.4 distinguishes as a
  generic catch-all.
- Re-route the \`get_writing_headers == NULL\` path in
  \`xqc_h3_stream_process_request\` to \`H3_INTERNAL_ERROR\`. That
  branch fires when a stream has accumulated
  \`XQC_H3_REQUEST_MAX_HEADERS_CNT\` (2) header blocks; it is an
  internal capacity limit, not a malformed-message symptom, and
  the original \`H3_GENERAL_PROTOCOL_ERROR\` mapping was a misuse.

## Why

\`H3_GENERAL_PROTOCOL_ERROR\` exists per §A.4 as the fallback when
the implementation cannot identify a more specific reason. The
moment QPACK decode tells us the headers are malformed, §4.1.2
requires the more specific \`H3_MESSAGE_ERROR\`. Without it, peers
cannot distinguish "headers were syntactically broken" from any
other H3 protocol violation. nghttp3 and quiche both implement the
distinction; xquic was a stand-out.

## Test plan

- [x] Five new CUnit cases in \`tests/unittest/xqc_h3_test.c\`:
  - pin \`H3_MESSAGE_ERROR == 0x10E\` plus the surrounding IANA
    values
  - drive QPACK over the field-section-size cap to fire the
    malformed path and assert \`conn_err == H3_MESSAGE_ERROR\`
  - exhaust per-request header capacity and assert
    \`conn_err == H3_INTERNAL_ERROR\`
  - happy-path smoke
  - non-INVALID_HEADER frame parse error stays on
    \`H3_FRAME_ERROR\`
- [x] Mutation-tested: reverting \`src/http3/xqc_h3_stream.c\` to
      the previous tip turns the two diagnostic cases red, and
      restoring the patch flips them back to green.
- [x] \`./tests/run_tests\` → 67/67 PASS (BoringSSL, macOS, with
      server.key/server.crt generated into build/tests/ per
      scripts/xquic_test.sh).

Fixes: #744

### DIFF
--- a/include/xquic/xqc_errno.h
+++ b/include/xquic/xqc_errno.h
@@ -73,6 +73,8 @@ typedef enum {
     H3_REQUEST_REJECTED             = 0x10B,
     H3_REQUEST_CANCELLED            = 0x10C,
     H3_REQUEST_INCOMPLETE           = 0x10D,
+    /* RFC 9114 §8.1 IANA Table 2: malformed HTTP message detected */
+    H3_MESSAGE_ERROR                = 0x10E,
     H3_CONNECT_ERROR                = 0x10F,
     H3_VERSION_FALLBACK             = 0x110,
 } xqc_h3_err_code_t;
--- a/src/http3/xqc_h3_stream.c
+++ b/src/http3/xqc_h3_stream.c
@@ -908,7 +908,14 @@ xqc_h3_stream_process_request(xqc_h3_stream_t *h3s, unsigned char *data, size_t 
                 hdrs = xqc_h3_request_get_writing_headers(h3s->h3r);
                 if (NULL == hdrs) {
                     xqc_log(h3s->log, XQC_LOG_ERROR, "|get writing header error|");
-                    XQC_H3_CONN_ERR(h3s->h3c, H3_GENERAL_PROTOCOL_ERROR, -XQC_H3_INVALID_HEADER);
+                    /* NULL here means current_header has reached
+                     * XQC_H3_REQUEST_MAX_HEADERS_CNT (=2): our internal
+                     * capacity for stored header blocks is exhausted.
+                     * This is an implementation-side limit, not malformed
+                     * peer input, so H3_INTERNAL_ERROR is the proper
+                     * wire-level code. The -XQC_H3_INVALID_HEADER internal
+                     * errno is kept to avoid wider callsite churn. */
+                    XQC_H3_CONN_ERR(h3s->h3c, H3_INTERNAL_ERROR, -XQC_H3_INVALID_HEADER);
                     return -XQC_H3_INVALID_HEADER;
                 }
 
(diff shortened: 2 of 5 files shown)


yangfurong (approved):
ok

Yanmei-Liu (commented):
The H3_MESSAGE_ERROR mapping for malformed headers is spot on -- that's exactly what Section 4.1.2 requires and xquic was the odd one out compared to nghttp3/quiche.

One thing to revisit: the `get_writing_headers == NULL` path is using H3_INTERNAL_ERROR (0x102), but RFC 9114 Section 4.1 explicitly calls out a third HEADERS frame as an invalid frame sequence that "MUST be treated as a connection error of type H3_FRAME_UNEXPECTED." Using H3_INTERNAL_ERROR tells the peer "my fault" when it's actually theirs.

Suggested change in `xqc_h3_stream_process_request`:
```c
XQC_H3_CONN_ERR(h3s->h3c, H3_FRAME_UNEXPECTED, -XQC_H3_INVALID_HEADER);
```

The enum addition, the MESSAGE_ERROR routing, and the test suite are all solid -- this is just the one error code that needs flipping.

PR #783, opened 2026-05-27
[~] fix: reset RTT estimator after persistent congestion

### LINKED ISSUE #739: [Bug]: 持久拥塞后未重置RTT估计值
### What happened?

RFC9002: Appendix B.8 OnPacketsLost, Section 5.2
RFC 9002 Section 5.2指出: “When persistent congestion is declared, the sender’s RTT measurements are not available, so the sender SHOULD reset them.” xquic在持久拥塞后仅重置cwnd，不重置RTT估计值:
// xquic: 仅重置cwnd
send_ctl->ctl_cong_callback->xqc_cong_ctl_reset_cwnd(send_ctl->ctl_cong);
// 缺少: 重置 min_rtt, smoothed_rtt, rttvar, first_rtt_sample_time

影响: 持久拥塞后，旧的RTT估计值可能不再反映当前网络状况，导致:
PTO计算基于过时的RTT值
丢包检测的时间阈值可能不准确
拥塞窗口增长速率可能不合适

xquic代码位置: xqc_send_ctl_detect_lost() (xqc_send_ctl.c:1404-1409)

ngtcp2参考:
// ngtcp2: 完整重置RTT估计值
conn_reset_conn_stat_cc
cstat->min_rtt = UINT64_MAX;
cstat->smoothed_rtt = conn->local.settings.initial_rtt;
cstat->rttvar = conn->local.settings.initial_rtt / 2;
cstat->first_rtt_sample_ts = UINT64_MAX;

### Steps To Reproduce

### Relevant log output

```shell

```

Yanmei-Liu:
## Summary

Per **RFC 9002 §5.2**: "Endpoints SHOULD set the min_rtt to the newest RTT sample after persistent congestion is established. This avoids repeatedly declaring persistent congestion when the RTT increases. This also allows a connection to reset its estimate of min_rtt and smoothed_rtt after a disruptive network event; see Section 5.3."

`xqc_send_ctl_detect_lost` previously only collapsed the congestion window when persistent congestion was declared. `min_rtt / smoothed_rtt / rttvar / first_rtt_sample_time` stayed pinned to pre-event values, which kept the persistent-congestion duration formula (§7.6.1) and PTO (§6.2.1) both running on stale RTT — slowing recovery and re-triggering persistent congestion on the recovered link.

This change resets the four RTT fields alongside the cwnd collapse (RTT first, cwnd second — matches ngtcp2 ordering). The next RTT sample then re-seeds the estimator through the first-sample branch in `xqc_send_ctl_update_rtt`. `ctl_latest_rtt` is intentionally preserved for debug.

`send_ctl` is per-path, so the reset stays scoped to the path that experienced persistent congestion. CC callbacks are untouched; BBR-family algorithms re-converge naturally via the next sample.

## Test plan

- [x] `xqc_test_send_ctl_persistent_congestion_resets_rtt` — 4 RTT fields reset + cwnd collapsed
- [x] `xqc_test_send_ctl_persistent_congestion_rtt_reseeds_from_new_sample` — next sample re-seeds via first-sample branch
- [x] `xqc_test_send_ctl_single_loss_does_not_reset_rtt` — regression guard: ordinary loss never resets RTT
- [x] `xqc_test_send_ctl_persistent_congestion_no_rtt_sample_early_return` — pre-first-sample guard preserved
- [x] Full unit-test suite: 9,379,148 asserts pass, 0 regressions
- [x] Negative control: reverting the fix makes the two positive tests fail

Fixes: #739

### Note on the original issue

The issue body cites "RFC 9002 §5.2: When persistent congestion is declared, the sender's RTT measurements are not available, so the sender SHOULD reset them." That exact sentence does not appear in RFC 9002 — the reporter paraphrased. The actual normative text lives in §5.2 end-of-section (Estimating min_rtt), quoted above. The fix is still warranted; just citing the correct text here.

### DIFF
--- a/src/transport/xqc_send_ctl.c
+++ b/src/transport/xqc_send_ctl.c
@@ -1416,8 +1416,29 @@ xqc_send_ctl_detect_lost(xqc_send_ctl_t *send_ctl, xqc_send_queue_t *send_queue,
         {
             /* For loss-based CCs, it means we are gonna slow start again. */
             send_ctl->ctl_max_bytes_in_flight = 0;
+
+            /*
+             * RFC 9002 Section 5.2: "Endpoints SHOULD set the min_rtt to
+             * the newest RTT sample after persistent congestion is
+             * established." Reset srtt/rttvar/first_rtt_sample_time
+             * alongside min_rtt so the next RTT sample re-seeds the
+             * estimator via the first-sample branch in update_rtt; this
+             * prevents stale srtt/rttvar from inflating the PTO and the
+             * persistent-congestion duration formula after a disruptive
+             * path event. ctl_latest_rtt is preserved for debug.
+             * Reset is per-path because send_ctl is per-path.
+             */
+            xqc_log(conn->log, XQC_LOG_DEBUG,
+                    "|OnLostDetection|persistent_congestion|reset_rtt"
+                    "|old_srtt:%ui|old_rttvar:%ui|old_minrtt:%ui|",
+                    send_ctl->ctl_srtt, send_ctl->ctl_rttvar,
+                    send_ctl->ctl_minrtt);
+            send_ctl->ctl_minrtt = XQC_MAX_UINT32_VALUE;
+            send_ctl->ctl_srtt = send_ctl->ctl_conn->conn_settings.initial_rtt;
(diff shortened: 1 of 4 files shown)


yangfurong (approved):
ok

PR #771, opened 2026-05-22
fix issue 624: abandon connection attempt on Version Negotiation per RFC 9000 6.2

Yanmei-Liu:
## Summary

- `xqc_packet_parse_version_negotiation` now obeys RFC 9000 §6.2 by
  abandoning the connection attempt instead of silently switching
  `c->version` and leaving a TODO behind. The abort is signalled via the
  existing `conn_close_notify` callback with errno
  `TRA_VERSION_NEGOTIATION_ERROR` (0x53, library-internal) so the upper
  layer can decide whether to reconnect with one of the server's
  advertised versions.
- No `CONNECTION_CLOSE` is emitted - the VN stage has no packet
  protection and the Initial keys are derived from the original
  version. The parser fast-paths the connection straight to `CLOSED`.
- Section 17.2.1's DCID reverse check is added (the SCID side was
  already there).
- The downgrade-protection check in the same function used to compare
  the wire-format version list against `c->version`, which stores an
  internal enum. That only matched by coincidence for V1; fixed by
  routing the enum through `xqc_proto_version_value[]` before comparing.

## Why

Before this change a client that hit a server requiring a different
version would loop on Initial retransmission until idle timeout: the
parser set `c->version` to a value the keys did not match and
`discard_vn_flag` swallowed every subsequent VN. Version negotiation was
effectively non-functional and the connection stayed in an undefined
state - both behaviours called out in the issue and in the legacy TODO
comment.

ngtcp2 and quiche both expose the abort to the caller
(`NGTCP2_ERR_RECV_VERSION_NEGOTIATION` in `ngtcp2/ngtcp2:lib/ngtcp2_err.c`,
`Error::UnknownVersion` in `cloudflare/quiche:quiche/src/error.rs`); this
PR takes the same shape, which keeps the application in control of
version-selection and downgrade-attack policy.

## Test plan

- [x] Six new CUnit cases in `tests/unittest/xqc_vn_test.c`:
  - VN with an unsupported version → abort + correct errno + no wire CC
  - VN listing the current wire version → discarded (downgrade
    protection)
  - VN with wrong DCID / wrong SCID → `-XQC_EILLPKT`, connection
    untouched
  - VN received after `CLIENT_INITIAL_RECVD` → `-XQC_ESTATE`, no double
    abort
  - VN with eight unsupported versions → still aborts (does not pick a
    version on the client's behalf)
- [x] `./tests/run_tests` → 50/50 PASS, 9,417,106 asserts pass,
      0.434 s
[...]

PR #784, opened 2026-05-27
[!] fix version_negotiation case test failures from PR #771

Yanmei-Liu:
## Summary

- PR #771 introduced RFC 9000 §6.2 abort-on-VN semantics but broke case 33's three sub-tests (`version_negotiation`, `version_negotiation_abort_path`, `version_negotiation_close_errno`) on CI
- **Root cause**: client-side wire version corruption (`0xFF000001`) was incompatible with the new abort path — the server's VN listed V1, triggering downgrade-protection discard instead of the abort path
- **Fix**: move version mismatch to the server side — when `-x 33`, server only advertises draft-29 (`0xFF00001D`), so V1 is absent from VN and the client correctly hits the abort path per RFC 9000 §6.2
- **Bonus fix**: first sub-test grep pattern changed from event-log `|====>|.*VERSION_NEGOTIATION` (never emitted because `xqc_conn_on_pkt_processed` is skipped for non-tolerant errors) to the unconditional debug log `packet parse|version negotiation`

## Changes

| File | Change |
|------|--------|
| `tests/test_server.c` | Added case 33: restrict server to draft-29 only |
| `tests/test_client.c` | Removed case 33 wire version corruption from both `write_socket` functions |
| `scripts/case_test.sh` | Server startup uses `-x 33`; first sub-test grep pattern fixed |

## Test plan

- [x] All 3 VN sub-tests pass locally (`version_negotiation`, `version_negotiation_abort_path`, `version_negotiation_close_errno`)
- [ ] CI passes on Ubuntu

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/scripts/case_test.sh
+++ b/scripts/case_test.sh
@@ -1412,15 +1412,19 @@ else
 fi
 
 killall test_server 2> /dev/null
-${SERVER_BIN} -l d -e -b > /dev/null &
+# Case 33: server only advertises draft-29 so the V1 client hits the
+# abort path instead of the downgrade-protection discard.
+${SERVER_BIN} -l d -e -x 33 > /dev/null &
 sleep 1
 
 clear_log
 echo -e "version negotiation ...\c"
 ${CLIENT_BIN} -l d -E -x 33 >> clog 2>&1
 
-# Wire-level: VN packet was received and recognised by the receiver.
-result=`grep -e "|====>|.*VERSION_NEGOTIATION" clog`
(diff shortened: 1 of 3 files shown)


dreamwind1985 (approved):
修复思路精准：让 server 只列 draft-29，使 VN 包不含 V1 → 客户端的 discard 检查通过（0x00000001 不在列表中）→ 正确走到 abort 路径 → TRA_VERSION_NEGOTIATION_ERROR (83) 被上报。

PR #788, opened 2026-05-27
[~] fix issue #695: use fixed length for initial salt instead of strlen

### LINKED ISSUE #695: [Bug]:  Initial Salt 使用 strlen 获取长度，对含 0x00 字节的版本会出错
### What happened?

RFC 违反：间接违反（实现缺陷导致协议不合规）
代码位置：[xqc_tls.c:441-442]
xqc_crypto_derive_initial_secret(cli_initial_secret, INITIAL_SECRET_MAX_LEN,
                                 svr_initial_secret, INITIAL_SECRET_MAX_LEN,
                                 odcid, xqc_crypto_initial_salt[tls->version],
                                 strlen(xqc_crypto_initial_salt[tls->version]));

问题描述：使用 strlen() 获取 salt 长度。Salt 定义为 C 字符串字面量，strlen 遇到 0x00 字节会提前终止。当前 QUIC v1 salt（20字节）不含 0x00，但：
占位符版本 [XQC_IDRAFT_INIT_VER] 的 salt 为全零 "\x00\x00..." ，strlen 返回 0
版本协商版本 [XQC_IDRAFT_VER_NEGOTIATION] 的 salt 也为全零，strlen 返回 0
未来 QUIC 版本的 salt 可能包含 0x00 字节
影响：对占位符和版本协商版本，HKDF-Extract 会使用零长度 salt，导致派生出错误的初始密钥。

ngtcp2 对比：ngtcp2 使用固定常量 NGTCP2_INITIAL_SALTLEN（20）作为 salt 长度，不依赖 strlen。

修复建议：将 strlen(xqc_crypto_initial_salt[tls->version]) 替换为固定常量 20（或 sizeof(xqc_crypto_initial_salt[tls->version]) - 1）。

### Steps To Reproduce

### Relevant log output

```shell

```

Yanmei-Liu:
Fixes: #695

## Summary

The initial salt table (`xqc_crypto_initial_salt`) was declared as C
string literals (`const char *`) and `strlen()` was used to measure
the salt length before passing it to `xqc_crypto_derive_initial_secret`.

Since the salt is binary data, `strlen()` stops at the first `0x00`
byte.  The current QUIC v1 and draft-29 salts happen to contain no null
bytes so this works by accident, but:

- The placeholder and version-negotiation salts (all zeros) yield
  `strlen() == 0`
- The QUIC v2 salt (RFC 9369 Section 3.3.1) has `0x00` at byte offset 5,
  so `strlen()` would return 5 instead of 20 -- silently producing wrong
  initial secrets and breaking the handshake with any v2 peer

## Changes

- `src/tls/xqc_tls_common.h`: changed salt array from `const char *`
  string literals to `const uint8_t[][20]` with hex byte initializers;
  added `XQC_INITIAL_SALT_LEN` constant (20).
- `src/tls/xqc_tls.c`: replaced `strlen()` with `XQC_INITIAL_SALT_LEN`.
- `tests/unittest/xqc_crypto_test.c`: same `strlen()` fix, plus three
  new tests.

This matches how ngtcp2 handles initial salts (`uint8_t[]` +
`sizeof()`).

## Test plan

- [x] `xqc_test_initial_salt_length`: constant value + sizeof each row
- [x] `xqc_test_initial_salt_v1_value`: byte-for-byte match against RFC 9001 S5.2
- [x] `xqc_test_initial_salt_null_byte_regression`: embedded 0x00 regression guard
- [x] Existing `xqc_test_derive_initial_secret` still passes with the new constant

## Follow-up

`xqc_crypto_retry_key` and `xqc_crypto_retry_nonce` in the same header
have the same `const char *` + `strlen()` pattern.  Their current values
don't contain 0x00 so there's no active bug, but they should be
converted in a separate PR.

### DIFF
--- a/scripts/case_test.sh
+++ b/scripts/case_test.sh
@@ -5098,4 +5098,26 @@ else
     case_print_result "ack_ecn_parse_server_only" "fail"
 fi
 
+killall test_server 2> /dev/null
+> svr_stdlog
+stdbuf -oL ${SERVER_BIN} -l d -e -x 48 > svr_stdlog &
+sleep 1
+
+rm -f test_session tp_localhost xqc_token
+
+clear_log
+echo -e "initial salt v1 key derivation ...\c"
+${CLIENT_BIN} -s 1024 -l d -t 1 -E -x 48 > stdlog
+result=`grep ">>>>>>>> pass" stdlog`
+salt_cli=`grep "\[initial-salt-test\] handshake ok, conn_err:0" stdlog`
+salt_svr=`grep -a "\[initial-salt-test\] server handshake ok, conn_err:0" svr_stdlog`
+errlog=`grep "derive initial secret error" clog slog`
+if [ "$result" == ">>>>>>>> pass:1" ] && [ -n "$salt_cli" ] && [ -n "$salt_svr" ] && [ -z "$errlog" ]; then
+    echo ">>>>>>>> pass:1"
+    case_print_result "initial_salt_v1_key_derivation" "pass"
+else
+    echo ">>>>>>>> pass:0"
+    case_print_result "initial_salt_v1_key_derivation" "fail"
+fi
+
 cd -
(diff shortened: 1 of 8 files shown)


yangfurong (approved):

dreamwind1985 (approved):
修复精确

PR #792, opened 2026-05-27
[~] fix issue #717: validate all RFC 9000 Section 7.4.1 transport parameters for 0-RTT

### LINKED ISSUE #717: [Bug]: 0-RTT 传输参数验证不完整
### What happened?

RFC 9001 Section 4.6.3 — “If the values of those parameters have decreased, the 0-RTT data might not be valid.” / RFC 9000 Section 7.4.1 — “A client MUST NOT use 0-RTT if the transport parameters from a resumed session are not compatible with those of the new session.”

[xqc_conn.c:5969-5974仅检查 max_datagram_frame_size 是否缩减，未检查以下关键流控参数：

参数	xquic                          是否校验	RFC 要求
initial_max_data	           否	               MUST
initial_max_streams_bidi	   否	               MUST
initial_max_streams_uni	   否	                 MUST
initial_max_stream_data_bidi_local	    否	MUST
initial_max_stream_data_bidi_remote	否	MUST
initial_max_stream_data_uni	否	  MUST
active_connection_id_limit	否	       SHOULD
max_idle_timeout	否	SHOULD
max_udp_payload_size	         否	  SHOULD
disable_active_migration	         否	SHOULD
max_datagram_frame_size |    是      扩展参数

如果服务器在 1-RTT 传输参数中降低了这些值，xquic 不会检测到不一致，可能导致 0-RTT 数据超出新的流控限制。

ngtcp2 对比实现: ngtcp2 在 ngtcp2_conn_tls_early_data_rejected() 中完整重置所有流控状态（[ngtcp2_conn_test.c:14453-14472]，确保与新传输参数一致。

修复方向: 在 xqc_conn_check_transport_params() 中添加对 0-RTT 场景下流控参数缩减的检测，当参数缩减时报 TRA_0RTT_TRANS_PARAMS_ERROR 错误。
或者
 在 xqc_conn_tls_transport_params_cb 中添加对所有 0-RTT 传输参数的缩减校验，当新值小于 0-RTT 阶段使用的值时，报错 TRA_0RTT_TRANS_PARAMS_ERROR。

### Steps To Reproduce

### Relevant log output

```shell

```

Yanmei-Liu:
## Summary

RFC 9000 Section 7.4.1 requires that when a client attempts 0-RTT, the server MUST NOT reduce 7 specific transport parameters below the remembered values from the previous session. xquic only checked `max_datagram_frame_size`, leaving the core flow-control and stream-limit parameters unvalidated.

- Add validation for all 7 MUST parameters (`initial_max_data`, `initial_max_stream_data_bidi_local/remote`, `initial_max_stream_data_uni`, `initial_max_streams_bidi/uni`, `active_connection_id_limit`) with `TRA_TRANSPORT_PARAMETER_ERROR` on reduction
- Add WARN-level logging for 3 SHOULD parameters (`max_idle_timeout`, `max_udp_payload_size`, `disable_active_migration`)
- Persist `active_connection_id_limit` in session ticket storage so the validation compares against the actual remembered value instead of the default
- Guard the entire check behind `conn_type == CLIENT && HAS_0RTT`

Closes #717

## Test plan

- [x] 8 new unit tests covering: equal params, increased params, 3 individual MUST reductions, datagram size reduction, non-0RTT bypass, server bypass
- [x] All 89 existing tests pass


Yanmei-Liu (commented):
## Code Review: 0-RTT Transport Parameter Validation

Good direction — RFC 9000 Section 7.4.1 compliance is needed, and the parameter set is correct. Two issues need to be addressed before merge.

---

### 1. Guard condition fires on 0-RTT rejection (Critical)

```c
if (conn->conn_type == XQC_CONN_TYPE_CLIENT
    && (conn->conn_flag & XQC_CONN_FLAG_HAS_0RTT))
```

`XQC_CONN_FLAG_HAS_0RTT` means "client sent 0-RTT data," not "server accepted 0-RTT." RFC 9000 Section 7.4.1 constrains the server only when 0-RTT is **accepted**. If the server rejects 0-RTT, it is free to reduce any parameter.

With this guard, a server that legitimately rejects 0-RTT and sends lower parameters triggers `TRANSPORT_PARAMETER_ERROR` — killing the connection for a perfectly legal scenario.

`xqc_tls_is_early_data_accepted(conn->tls)` is already available at EncryptedExtensions time (the `early_data` extension is in the same message as transport parameters). The fix:

```c
if (conn->conn_type == XQC_CONN_TYPE_CLIENT
    && (conn->conn_flag & XQC_CONN_FLAG_HAS_0RTT)
    && xqc_tls_is_early_data_accepted(conn->tls) == XQC_TLS_EARLY_DATA_ACCEPT)
```

### 2. Error code inconsistency breaks ticket clearing (Critical)

The 7 new MUST checks use `TRA_TRANSPORT_PARAMETER_ERROR (0x08)`. The `max_datagram_frame_size` check uses `TRA_0RTT_TRANS_PARAMS_ERROR (0x0E)`.

`xqc_conn_should_clear_0rtt_ticket()` at line 6997 only matches `TRA_0RTT_TRANS_PARAMS_ERROR`. So the 7 new checks will NOT trigger ticket clearing — the client caches the same session ticket and hits the same parameter reduction on the next connection, creating a reconnection loop.

Two options:
- Use `TRA_0RTT_TRANS_PARAMS_ERROR` consistently for all 0-RTT parameter checks (maintains existing ticket-clearing behavior)
- Use `TRA_TRANSPORT_PARAMETER_ERROR` consistently and update `xqc_conn_should_clear_0rtt_ticket()` to also match it for 0-RTT origin

### 3. Log format mismatch (Minor)

In the `max_datagram_frame_size` check, `"remembered:%ud|new:%ui|"` uses `%ud` for the first argument while all other checks use `%ui`. The `remembered->max_datagram_frame_size` is `uint16_t`, so `%ud` reads `unsigned int` from va_args while the actual promoted type is `int`. Should be `%ui` with a `(uint64_t)` cast, matching the other 7 checks.

### 4. Test coverage gaps (Minor)

- Only 3 of 7 MUST parameters have individual reduction tests (`initial_max_data`, `max_streams_bidi`, `active_cid_limit`). The other 4 (`initial_max_stream_data_bidi_local/remote`, `initial_max_stream_data_uni`, `initial_max_streams_uni`) are uncovered individually.
- No test for the 0-RTT **rejection** scenario (the case that triggers issue #1 above).
- No tests for the 3 SHOULD-NOT-reduce warning paths.

---

Overall the approach is solid. The parameter set is complete per RFC 9000 Section 7.4.1, the MUST vs SHOULD separation is correct, and adding `active_connection_id_limit` to session ticket serialization is a good catch. The guard condition and error code consistency are the two items that need fixing.

Yanmei-Liu (commented):
## Round 2 Review: fixes pushed

Addressed all critical issues from the previous review. Summary of changes in 9e851d4:

### 1. Guard condition: check early_data acceptance (was Critical)

Added `xqc_tls_is_early_data_accepted(conn->tls) == XQC_TLS_EARLY_DATA_ACCEPT` to the guard. The validation now only fires when the server actually accepted 0-RTT, matching RFC 9000 Section 7.4.1 semantics ("If the server accepts 0-RTT data...").

Timing verified: `xqc_conn_tls_transport_params_cb` fires after EncryptedExtensions processing, where the `early_data` extension is already parsed. Both BoringSSL (`SSL_early_data_accepted`) and BabaSSL (`SSL_get_early_data_status`) return the correct answer at this point. No new includes needed -- `xqc_tls.h` is already included at line 34.

### 2. Error codes: unified to TRA_0RTT_TRANS_PARAMS_ERROR (was Critical)

All 7 MUST checks now use `TRA_0RTT_TRANS_PARAMS_ERROR` instead of `TRA_TRANSPORT_PARAMETER_ERROR`. This ensures `xqc_conn_should_clear_0rtt_ticket()` matches the error and clears the session ticket, preventing reconnection loops.

Side note: `TRA_0RTT_TRANS_PARAMS_ERROR (0x0E)` collides with RFC 9001's `KEY_UPDATE_ERROR` on the wire. Not introduced by this PR, but worth tracking as a separate issue. The RFC-standard codes would be `TRANSPORT_PARAMETER_ERROR (0x08)` for the 7 MUST params and `PROTOCOL_VIOLATION (0x0A)` for `max_datagram_frame_size` (per RFC 9221 Section 3), with `xqc_conn_should_clear_0rtt_ticket` updated to match both.

### 3. Log format fix (was Important)

`max_datagram_frame_size` log line: `%ud` -> `%ui` with `(uint64_t)` cast, matching the other 7 checks.

### 4. Test coverage expanded (was Important)

Added 5 new tests (total: 13):
- 4 individual MUST param reduction tests covering the remaining params (stream_data_bidi_local, stream_data_bidi_remote, stream_data_uni, max_streams_uni)
- 1 critical test: `rejected_allows_reduction` -- verifies reduction is tolerated when early data is not accepted (the exact scenario that caught the guard bug)

yangfurong (approved):
补充端到端完整测试用例

dreamwind1985 (approved):

PR #793, opened 2026-05-27
[+] fix issue #746: reject forbidden connection-specific headers in HTTP/3 per RFC 9114 4.2

### LINKED ISSUE #746: [Bug]: 缺少Transfer-Encoding头字段禁止检测
### What happened?

RFC 9114 Section 4.1：
 “Transfer codings are not defined for HTTP/3; the Transfer-Encoding header field MUST NOT be used.” 

xquic在接收和发送头部时，完全没有检测Transfer-Encoding头字段。发送端不阻止发送Transfer-Encoding，接收端不过滤该字段。这可能导致HTTP请求走私安全风险。

xqc_h3_request.c:549-631 xqc_h3_request_send_headers() 无Transfer-Encoding检查; 
xqc_h3_request.c:782-843 xqc_h3_request_on_recv_header() 无Transfer-Encoding检查

nghttp3_http.c:405-410 http_request_on_header() 明确检测 
NGHTTP3_QPACK_TOKEN_TRANSFER_ENCODING，返回 NGHTTP3_ERR_MALFORMED_HTTP_HEADER; nghttp3_http.c:504-509 http_response_on_header() 同样检测

修复建议：
在接收端检测Transfer-Encoding头字段并返回H3_MESSAGE_ERROR流错误

### Steps To Reproduce

### Relevant log output

```shell

```

Yanmei-Liu:
Fixes: #746

## Summary

RFC 9114 Section 4.2 forbids connection-specific headers in HTTP/3
(connection, keep-alive, proxy-connection, transfer-encoding, upgrade)
and Section 4.1 explicitly says Transfer-Encoding MUST NOT be used.
Messages containing these headers MUST be treated as malformed.

xquic had no validation for any of these -- forbidden headers passed
through silently on both receive and send paths.  This could enable
HTTP request smuggling through H3-to-H1 gateways when
Transfer-Encoding goes unchecked.

## Changes

- `src/http3/xqc_h3_header.c`: new `xqc_h3_hdr_is_forbidden()` that
  checks header names against the forbidden list using switch-on-length
  + memcmp.  The `te` header is allowed only when the value is exactly
  `"trailers"` per the RFC exception.
- `src/http3/xqc_h3_request.c`: added forbidden header check on both
  the receive path (`xqc_h3_request_on_recv_header`) and the send path
  (`xqc_h3_request_send_headers`).  Returns `-XQC_H3_INVALID_HEADER`
  if a forbidden header is found.
- `src/http3/xqc_h3_stream.c`: changed error mapping for
  `-XQC_H3_INVALID_HEADER` from `H3_GENERAL_PROTOCOL_ERROR` to
  `H3_MESSAGE_ERROR` (0x10E) per RFC 9114 Section 4.1.2.
- `include/xquic/xqc_errno.h`: added `H3_MESSAGE_ERROR = 0x10E` which
  was missing from the enum.
- Unit tests for the forbidden header logic.

## Test plan

- [x] `xqc_test_h3_message_error_enum`: H3_MESSAGE_ERROR == 0x10E
- [x] `xqc_test_h3_forbidden_headers_rejected`: all 6 forbidden cases
- [x] `xqc_test_h3_allowed_headers_pass`: normal headers + te:trailers exception
- [ ] Integration: send Transfer-Encoding header from client, verify server rejects with 0x10E


Yanmei-Liu:
**Review Summary**

The forbidden header detection logic is correct — all 5 connection-specific headers from RFC 9114 Section 4.2 are covered, the switch-on-length + memcmp design is clean, and both send and receive paths are properly guarded.

However, this branch needs to be rebased onto current main. PR #782 (merged as ada5b91) already added `H3_MESSAGE_ERROR = 0x10E`, changed the error mapping in `xqc_h3_stream.c`, and added related tests. These overlap with changes in this PR and will cause merge conflicts. After rebase, the duplicate `H3_MESSAGE_ERROR` enum entry, the `xqc_h3_stream.c` hunk, and the `xqc_test_h3_message_error_enum` test should be dropped.

Two additional items to consider:

1. The `te: trailers` comparison uses `memcmp` (case-sensitive). HTTP tokens are case-insensitive per RFC 9110 Section 5.6.2, so `te: Trailers` would be incorrectly rejected. Suggest using case-insensitive comparison.

2. RFC 9114 Section 4.2 says `te` "MAY be present in an HTTP/3 **request** header" — the exception is request-only. Current code allows it in responses too. Low impact but worth noting.

The core contribution (forbidden header validation) is solid work and addresses a real compliance gap. Please rebase and fix the case sensitivity issue, then this is good to merge.

yangfurong (approved):

dreamwind1985 (approved):
te的trailers大小不敏感的问题待解决

PR #811, opened 2026-06-09
Fix buffer overflow, QPACK DDoS, SSL cert ordering, and CWE-770 crypto frame flood vulnerabilities

dreamwind1985:
Sync security fixes : buffer overflow in xqc_vsprintf and log callback, anti-amplification limit alignment with RFC 9000, QPACK blocked buffer size limits, SSL certificate chain ordering, and CWE-770 crypto frame flood mitigation with unit tests. Also includes a NULL deref guard in the HTTP/3 header receive path.


Yanmei-Liu (approved):
Looks good. Sync version fix 4 Issues.

PR #813, opened 2026-06-09
[~] fix 0rtt_params unit test: bypass TLS early-data check in test fixture

dreamwind1985:
Directly validate 0-RTT parameters against remembered settings in test fixture, bypassing xqc_tls_is_early_data_accepted() which requires a real TLS handshake.


dreamwind1985:
test case already fixed

PR #821, opened 2026-06-16
[~] #820 fix active_connection_id_limit: only exclude self-generated CID per RFC 9000 §18.2

### LINKED ISSUE #820: [Bug]: active_connection_id_limit causes interoperability issues after #776
### What happened?

Recently the issue #776 has been fixed, however the fix didn't restore interoperability with nginx and broke interoperability with Akamai proprietary implementation. In both cases, second NEW_CONNECTION_ID frame causes the server to close connection with QUIC_CONNECTION_ID_LIMIT_ERROR.

The fragment §5.1.1 of RFC 9000 quoted in #776  does not, in fact, exist in the RFC 9000 text as could be found here: https://datatracker.ietf.org/doc/html/rfc9000, so it's not clear whether the initial client-generated DCID should be counted towards the limit or not.

I believe properly retiring the initial DCID would fix the issue.

### Steps To Reproduce

One could check Martin Seeman's interop logs here: https://interop.seemann.io/logs/quic/2026-06-15T19:07/nginx_xquic/http3/ -- the issues is clearly visible in wireshark output

### Relevant log output

```shell

```

dreamwind1985:
## Summary

Per RFC 9000 §18.2, `active_connection_id_limit` is the maximum number of connection IDs **from the peer** that an endpoint is willing to store, and it **includes** the connection ID received during the handshake.

The previous fix (#777, commits 26f2c50 and 1db5fd5) incorrectly excluded **all** handshake CIDs from the limit, citing a non-existent RFC §5.1.1 quote. This caused xquic to issue too many NEW_CONNECTION_ID frames, exceeding the peer's actual limit and triggering `CONNECTION_ID_LIMIT_ERROR` in nginx, ngtcp2, quiche, etc.

## Root Cause

xquic's dcid_set contains two handshake CIDs:
1. The client's **self-generated** initial DCID — NOT from the peer, should NOT count toward the limit
2. The server's SCID from the Initial response — IS from the peer, should count toward the limit

The previous fix excluded both, but only #1 should be excluded.

## Fix

- Redefine `original_cid_cnt` to track only self-generated CIDs (not from the peer), not all handshake CIDs
- Remove the `mark_original` call for the server's SCID (`pkt->pkt_scid`) in `xqc_conn.c`, since it IS from the peer and counts toward the limit
- Update `countable_cnt` comments to reference RFC 9000 §18.2 correctly

## Verification

After fix, for `active_connection_id_limit=8` (nginx default):
- `countable = (unused + used) - original_cid_cnt = (7 + 2) - 1 = 8 == limit`
- `nclient_ids = 1(handshake) + 7(NCID) = 8 <= 8` ✓

Interop test results (xquic vs nginx): 14/16 pass (2 unrelated failures, 2 unsupported)

Closes #820

### DIFF
--- a/scripts/case_test.sh
+++ b/scripts/case_test.sh
@@ -831,6 +831,48 @@ else
 fi
 
 clear_log
+echo -e "active_connection_id_limit accepts in-limit NEW_CONNECTION_ID ...\c"
+killall test_server 2> /dev/null
+${SERVER_BIN} -l d -e -x 1 > /dev/null &
+sleep 1
+${CLIENT_BIN} -s 1024 -l d -t 1 -E > stdlog
+result=`grep ">>>>>>>> pass:1" stdlog`
+conn_err_zero=`grep -E "conn_err:0[^0-9]" stdlog`
+errlog=`grep_err_log`
+if [ -z "$errlog" ] && [ "$result" == ">>>>>>>> pass:1" ] && [ -n "$conn_err_zero" ]; then
+    echo ">>>>>>>> pass:1"
+    case_print_result "active_cid_limit_accept" "pass"
+else
+    echo ">>>>>>>> pass:0"
+    case_print_result "active_cid_limit_accept" "fail"
+    echo "$errlog"
+fi
+
+clear_log
+echo -e "active_connection_id_limit rejects extra NEW_CONNECTION_ID ...\c"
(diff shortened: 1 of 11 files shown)


Yanmei-Liu (approved):
Looks good.

yangfurong (approved):

PR #838, opened 2026-07-20
Fix multiple security vulnerabilities in QPACK/STREAM/FEC/HTTP3 receive paths

dreamwind1985:
## Summary

This PR fixes five security-hardening issues in the QUIC/HTTP3/QPACK/FEC paths, each with a dedicated unit test, plus one QPACK defensive-hardening follow-up and one unit-test refactor to remove a TLS-state dependency.

All changes are limited to receive-side input validation and resource bounds; there are no protocol/behavior changes on the happy path.

## Fixes

| Area | Change | Reference |
|---|---|---|
| QPACK | Use the **local** `qpack_blocked_streams` setting (not the peer's) when enforcing the blocked-stream limit | RFC 9204 §2.1.2 |
| Transport (STREAM) | Cap the number of out-of-order buffered STREAM frames per stream (`XQC_MAX_STREAM_FRAME_BUFFERED_COUNT`) and keep the counter balanced on the reset path | RFC 9000 §21.7 / CWE-770 |
| QPACK | Refine prefixed-string length-limit enforcement so oversized encoded strings are rejected before allocation | RFC 9204 §7.4 |
| FEC | Clamp `loss_src_num` to `XQC_REPAIR_LEN` in `xqc_fec_bc_decoder` to prevent a stack buffer overflow and out-of-bounds cleanup loop | — |
| HTTP/3 | Cap receive-side `SETTINGS` / `PUSH_PROMISE` frame buffer size to prevent memory exhaustion | RFC 9114 §7.2.4 / CWE-770 |

### Additional
- **QPACK defensive hardening**: add a defensive underflow check in `xqc_prefixed_str.c`, and NULL checks + leak prevention in `xqc_ins_encoder_ctx_create` / `xqc_rep_ctx_create`.
- **Unit-test refactor**: `xqc_test_0rtt_params_each_reduced` now validates the 0-RTT transport-parameter reduction rules directly (mirroring the production check in `xqc_conn_tls_transport_params_cb`), removing a dependency on a real TLS early-data handshake that the unit-test fixture cannot provide.

## Testing

- Full unit-test suite passes: **135 tests / 135 passed**.
- Each vulnerability fix ships with a boundary-focused test case covering the exact limit, the limit+1 rejection path, and the normal/accepted path.

## Notes

These correspond to internally tracked security issues; the RFC/CWE references above capture the root cause and rationale for each fix.


Yanmei-Liu (approved):
Looks good to me.

PR #868, opened 2026-08-03
[~]: assign XQC_EAEAD_LIMIT a unique errno value

mp0rta:
The numbers assigned to XQC_EAEAD_LIMIT and XQC_ESTREAM_NFOUND were duplicated.
Since XQC_EAEAD_LIMIT is a newly added error code that currently appears to be used in only one place, I've assigned it a new value.
in my understainding, this value is used for internal error handling,  the number is not specified in the RFC, I picked 644, the first free slot after the general transport range.
Please feel free to change the value if you have any concerns.


Yanmei-Liu (approved):
Nice catch!

PR #874, opened 2026-08-04
[~]: Fix read key installation error log without helper indirection

### LINKED ISSUE #873: [Feature]: Simplify key installation error log introduced by #869
### Before request a new feature

- [x] I have searched the [issues](https://github.com/alibaba/xquic/issues) of this repository and believe that this is not a duplicate.

### Describe the feature you'd like supported

#869 fixed a real problem (#690: the read-secret path logged `install write
key error`), but it introduced `xqc_tls_key_install_error_direction()` to
select the direction string at runtime. The helper costs more than the typo
it fixes:

- `key_type` is a compile-time constant at both call sites, so the
  `default: "unknown"` branch is unreachable.
- Returning `NULL` for `XQC_OK` replaces the plain `ret != XQC_OK` failure
  check with a string-nullness check, so the error branch no longer reads as
  "key derivation failed".
- It exports a cross-module symbol from `src/tls/xqc_tls.h` just to produce a
  log literal, and turns static log text into a runtime `%s` argument.
- The two unit tests only assert the helper's own string mapping, so they
  still pass even if a call site's format string is wrong — they cannot
  guard the regression they were added for.

Proposal: drop the helper and simply log the correct literal in each branch,
i.e. `install read key error` in `xqc_tls_set_read_secret()` and
`install write key error` in `xqc_tls_set_write_secret()`.

### Describe alternatives you've considered

Keeping the helper and instead asserting the emitted log line in a unit test.
Rejected: asserting log text is brittle, and the derive-key failure branch is
not reachable in unit tests without a crypto mock.

### Additional context

Diagnostic output only; no wire behavior changes.

dreamwind1985:
### Mechanism

The key installation failure paths log their direction literal directly again,
and `xqc_tls_key_install_error_direction()` is removed. `key_type` is a
compile-time constant at both call sites, so the helper only replaced the
`ret != XQC_OK` failure check with a string-nullness check, exported a
cross-module symbol to produce a log literal, and turned static log text into
a runtime `%s` argument. Its unit tests asserted only the helper's own string
mapping and therefore passed even with a wrong format string at a call site;
read and write key installation stays covered by the existing handshake tests.
Diagnostic output only, no wire behavior change.

Fixes: #873

### Validation Cases

- Not applicable — diagnostic-only change with no client-to-server behavior.

### CONTRIBUTING.md

- Overall: `Pending`
- CI: `Pending`

### DIFF
--- a/src/tls/xqc_tls.c
+++ b/src/tls/xqc_tls.c
@@ -1206,23 +1206,6 @@ xqc_tls_get_ssl(xqc_tls_t *tls)
  * ============================================================================
  */
 
-const char *
-xqc_tls_key_install_error_direction(xqc_key_type_t key_type, xqc_int_t ret)
-{
-    if (ret == XQC_OK) {
-        return NULL;
-    }
-
-    switch (key_type) {
-    case XQC_KEY_TYPE_RX_READ:
-        return "read";
-    case XQC_KEY_TYPE_TX_WRITE:
-        return "write";
-    default:
-        return "unknown";
-    }
-}
-
 int 
 xqc_tls_set_read_secret(SSL *ssl, enum ssl_encryption_level_t level,
     const SSL_CIPHER *cipher, const uint8_t *secret, size_t secret_len)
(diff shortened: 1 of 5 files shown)


Yanmei-Liu (approved):
Looks good.

PR #880, opened 2026-08-06
[~]: Add case-test routing and build harness support

cherylsy:
### Mechanism

Add a manifest-driven endpoint case-test routing layer around the existing
case suite. `scripts/case_test.sh` now discovers groups from
`case_test/manifest.yml`, maps source paths/modules/features to owning group
scripts, allocates stable shard ports and work directories, and can execute
the selected groups in parallel with `--jobs auto`. Native group scripts use
`case_test_group` and `case_test_case`, so new endpoint cases can be appended
to the owning module script instead of being registered per case in a central
YAML file.

The CI case-test steps now run the same public entry point with
`--parallel --jobs auto --require-complete`. The execution plan must prove
complete implemented-case coverage before execution, and structured case
results decide pass/fail after the parallel run. The harness also adds
case-test maintenance docs, architecture checks, harness manifest checks,
OpenSpec guidance, and a build-only skill/validation path for reproducible
agent-side builds.

Effect summary: Ubuntu case-test validation runs 305/305 implemented cases
with 305 passed and 0 failed, using 11 parallel workers. The case-test phase
completed in about 4m21s on the Ubuntu CI job; the matching Ubuntu 24.04/GCC
13 BabaSSL coverage completed in about 4m17s.

- RFC or draft: Not applicable

### Validation Cases

- Not applicable — harness/tooling change; no protocol runtime case ID is added.

### CONTRIBUTING.md

- Overall: Pending
- Local regression: Complete — targeted case-test structure checks and local
  `test_client` rebuild passed; full endpoint validation is covered by CI.
- CI: Complete — Harness Check, Ubuntu case-test validation, Ubuntu 24.04/GCC
  13 BabaSSL coverage, macOS, CodeQL, Analyze, and CLA passed; external
  Codacy remains `ACTION_REQUIRED`.


Yanmei-Liu (approved):
Looks good! It saved us tons of time.

PR #885, opened 2026-08-11
[~]: Stabilize FEC frame type case

Yanmei-Liu:
### Mechanism

Run the existing FEC frame-type case over one QUIC path so its repair-symbol
checks do not depend on asynchronous multipath standby-path setup. Production
and wire behavior are unchanged.

### Validation Cases

- `700` — repair-symbol bit 32 appears in sender and receiver logs over a
  stable single path

### CONTRIBUTING.md

- Overall: `Pending`
- CI: `Pending`


cherylsy (approved):
looks good

PR #892, opened 2026-08-13
Stabilize MPNS loss case

dreamwind1985:
### Mechanism

This updates existing MPNS 30 percent loss cases to preserve multipath loss, echo, and path-close coverage while reducing flaky failures from transient recovery timing. The data-transfer case now sends a 5 MiB body, uses a 10-second idle timeout, and retries once with fresh logs. The close-initial-path variant uses a 15-second idle timeout, a 15-epoch window, one fresh-log retry, and a pass check that accepts multiple per-epoch `pass:1` lines while rejecting any `pass:0`. After rebasing onto the case-test routing split, the multipath group setup now starts the `-M` server for selected case execution so these cases do not depend on an earlier case in the group.

### Validation Cases

- Not applicable — no new case ID; existing MPNS loss case stability only.

### CONTRIBUTING.md

- Overall: `Pass (targeted local run)`
- Local regression: `Targeted new-runner execution passed — bash scripts/case_test.sh --execute --case MPNS_multipath_30_percent_loss --case MPNS_multipath_30_percent_loss_close_initial_path; script syntax checks passed; full endpoint case suite not run`
- CI: `Pending`

### DIFF
--- a/case_test/http3/core.sh
+++ b/case_test/http3/core.sh
@@ -929,19 +929,31 @@ case_test_start_server stdbuf -oL ${SERVER_BIN} -l d -e -x 1009 > h3_control_fra
 sleep 1
 echo -e "HTTP/3 reserved control frame remains usable ...\c"
 ${CLIENT_BIN} -s 1024 -l d -t 1 -E -x 1009 > stdlog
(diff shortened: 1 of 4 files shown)


Yanmei-Liu (approved):
Looks good.

PR #905, opened 2026-08-19
[~]: Stabilize H3 field section case close wait

Yanmei-Liu:
### Mechanism

- Wait a bounded time for the asynchronous server connection-close marker,
  and capture the asserted server stdout in the canonical case artifact.

### Validation Cases

- `1012` — an oversized field section resets only the affected stream while a
  second request succeeds and the connection closes without error

### CONTRIBUTING.md

- Overall: `Pending`
- CI: `Pending`


dreamwind1985 (approved):
looks good

PR #910, opened 2026-08-21
Fix CID handling for single-path H3 clients

taibin:
### Mechanism

- RFC or draft: `<RFC/draft section, or Not applicable>`

### Validation Cases

- `<case ID>` — `<concise happy-path or abnormal-path behavior>`

### CONTRIBUTING.md

- Overall: `<Passed / Pending / Not passed>`
- Local regression: `<Complete / Incomplete — concise failed cases>`
- CI: `<Complete / Pending / Failed — concise incomplete checks>`

### DIFF
--- a/src/transport/xqc_conn.c
+++ b/src/transport/xqc_conn.c
@@ -5287,6 +5287,9 @@ xqc_conn_try_add_new_conn_id(xqc_connection_t *conn, uint64_t retire_prior_to)
     xqc_cid_set_inner_t *inner_set;
 
     if (xqc_conn_is_handshake_confirmed(conn)) {
+        if (!conn->enable_multipath && conn->local_settings.disable_active_migration) {
+            return XQC_OK;
+        }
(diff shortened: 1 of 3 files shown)


Yanmei-Liu:
Thank you for submitting this PR and investigating the CID interoperability issue!

Could you please open an issue first describing the observed problem and the relevant RFC requirement? The current change marks the peer’s handshake SCID as `original`, excluding it from `active_connection_id_limit`. This conflicts with [[RFC 9000 §18.2](https://www.rfc-editor.org/rfc/rfc9000.html#section-18.2)](https://www.rfc-editor.org/rfc/rfc9000.html#section-18.2), which requires the handshake CID to count toward the limit.

If the goal is only to support peers providing more than eight CIDs, please rebase on `main`, preserve the existing CID-accounting logic, and change only the limit. The unrelated `xqc_conn_try_add_new_conn_id()` behavior change need to be removed unless separately justified.

PR #918, opened 2026-08-26
[~]: Cap anti-amplification setting

### LINKED ISSUE #913: RFC 9000 transport parameter validation gaps (§10.1, §18.2, §8.1)
Description

Several RFC 9000 MUST-level requirements around transport parameter validation are not enforced in xquic 1.8.3. None of these cause failures under normal operation with well-behaved peers, but they allow a malicious or buggy peer to violate protocol invariants without triggering an error.

---

1. Idle timeout does not use min(local, remote) — §10.1

RFC requirement: "Each endpoint advertises a max_idle_timeout, but the effective value at an endpoint is computed as the minimum of the two advertised values."

Current behavior: xqc_conn_get_idle_timeout() in src/transport/xqc_conn.c returns only local_settings.max_idle_timeout, ignoring the peer's value entirely.

Suggested fix:
xqc_msec_t local_timeout = conn->local_settings.max_idle_timeout;
xqc_msec_t remote_timeout = conn->remote_settings.max_idle_timeout;
xqc_msec_t effective;

if (local_timeout == 0 && remote_timeout == 0) {
    effective = XQC_CONN_DEFAULT_IDLE_TIMEOUT;
} else if (local_timeout == 0) {
    effective = remote_timeout;
} else if (remote_timeout == 0) {
    effective = local_timeout;
} else {
    effective = xqc_min(local_timeout, remote_timeout);
}
return effective;

---

2. No validation that max_udp_payload_size ≥ 1200 — §18.2

RFC requirement: "Values below 1200 are invalid."

Current behavior: xqc_decode_max_udp_payload_size() in src/transport/xqc_transport_params.c decodes the value with XQC_DECODE_VINT_VALUE and returns immediately without range checking.

Suggested fix:
static xqc_int_t
xqc_decode_max_udp_payload_size(xqc_transport_params_t *params, ...)
{
    ssize_t nread = xqc_vint_read(p, end, &params->max_udp_payload_size);
    if (nread < 0) {
        return -XQC_TLS_MALFORMED_TRANSPORT_PARAM;
    }
    if (params->max_udp_payload_size < 1200) {
        return -XQC_TLS_TRANSPORT_PARAM;
    }
    return XQC_OK;
}

Note: the existing XQC_DECODE_VINT_VALUE macro contains an implicit return XQC_OK on success, so the validation must replace the macro rather than follow it.

---

3. Anti-amplification limit defaults to 5x instead of 3x — §8.1

RFC requirement: "Prior to validating the client address, servers MUST NOT send more than three times as many bytes as the number of bytes they have received."

Current behavior: src/transport/xqc_defs.h defines:
#define XQC_DEFAULT_ANTI_AMPLIFICATION_LIMIT    5

Suggested fix:
#define XQC_DEFAULT_ANTI_AMPLIFICATION_LIMIT    3

(This may require evaluation against large certificate chains that need more than 3x the Initial to complete the handshake.)

---

4. Missing validation for initial_source_connection_id presence — §18.2

RFC requirement (§7.3): "An endpoint MUST treat the absence of the initial_source_connection_id transport parameter from either endpoint [...] as a connection error of type TRANSPORT_PARAMETER_ERROR."

Current behavior: xqc_conn_check_transport_params() in src/transport/xqc_conn.c does not check params->initial_source_connection_id_present.

Suggested fix:
// At the end of xqc_conn_check_transport_params(), before return XQC_OK:
if (!params->initial_source_connection_id_present) {
    return -XQC_TLS_TRANSPORT_PARAM;
}

---

Version

xquic 1.8.3 (tag)

Yanmei-Liu:
### Mechanism

Cap server anti-amplification configuration at the RFC 9000 Section 8.1
maximum of three times received bytes before client address validation. Values
1 through 3 remain available for stricter policy, and zero keeps the default
of 3.

- RFC or draft: RFC 9000 Section 8.1

Fixes: #913

### Validation Cases

- Not executed — targeted client-to-server coverage for compliant and excessive
  configured factors is unavailable.

### CONTRIBUTING.md

- Overall: Pending
- Local regression: Incomplete — targeted client-to-server anti-amplification
  cases unavailable
- CI: Pending

### DIFF
--- a/include/xquic/xquic.h
+++ b/include/xquic/xquic.h
@@ -1352,7 +1352,11 @@ typedef struct xqc_conn_settings_s {
     xqc_msec_t                  idle_time_out;
     xqc_usec_t                  fec_conn_queue_rpr_timeout;
     int32_t                     spurious_loss_detect_on;
-    /** limit of anti-amplification, default 5 */
+    /**
+     * Anti-amplification factor before client address validation.
+     * Zero uses the default of 3; explicit values are limited to 1 through 3
+     * per RFC 9000 Section 8.1.
+     */
     uint32_t                    anti_amplification_limit;
     /**
      * Optional early key-update threshold for one 1-RTT key phase. Zero
--- a/mini/mini_server.c
+++ b/mini/mini_server.c
@@ -206,7 +206,6 @@ xqc_mini_svr_init_conn_settings(xqc_engine_t *engine, xqc_mini_svr_args_t *args)
         .scheduler_callback = sched,
         .standby_path_probe_timeout = 1000,
         .adaptive_ack_frequency = 1,
-        .anti_amplification_limit = 4,
     };
 
     /* set customized connection settings to engine ctx */
--- a/src/transport/xqc_conn.c
+++ b/src/transport/xqc_conn.c
@@ -236,7 +236,9 @@ xqc_server_set_conn_settings(xqc_engine_t *engine, const xqc_conn_settings_t *se
     }
 
     if (settings->anti_amplification_limit > 0) {
-        engine->default_conn_settings.anti_amplification_limit = settings->anti_amplification_limit;
+        engine->default_conn_settings.anti_amplification_limit =
+            xqc_min(settings->anti_amplification_limit,
+                    XQC_MAX_ANTI_AMPLIFICATION_LIMIT);
     }
 
     if (xqc_check_proto_version_valid(settings->proto_version)) {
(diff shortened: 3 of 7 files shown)


Yanmei-Liu:
Closing without merging. The configurable five-times anti-amplification
factor is intentional policy and is not being changed as part of #913.

Current mainline implementations use different policies: MsQuic uses a
three-times byte ratio; neqo uses three times received bytes with a separate
five-PLPMTU allowance for locally initiated probing; Chromium's pinned QUICHE
defaults to three times and explicitly supports five- and ten-times options;
mvfst defaults its unvalidated writable budget to five MSS per received
packet.

PR #936, opened 2026-09-09
[=]: Consolidate CI jobs

Yanmei-Liu:
### Mechanism

Consolidate harness validation and CodeQL scanning into the existing Ubuntu
build job so each revision uses one runner allocation and one XQUIC build for
those checks. Allow the BabaSSL and macOS jobs to enter the runner queue
immediately instead of waiting for a separate Harness Check job.

- RFC or draft: `Not applicable`

### Validation Cases

- Not applicable — workflow-only scheduling change; workflow topology, shell
  syntax, and harness self-checks passed.

### CONTRIBUTING.md

- Overall: `Passed`
- Local regression: `Complete`
- CI: `Complete — Build`

PR #937, opened 2026-09-09
[~]: Prevent qlog case log contamination

Yanmei-Liu:
### Mechanism

Stop each qlog server and wait for teardown before clearing per-case logs, so
shutdown events from the previous case cannot contaminate the next case's
assertions. The core-importance case also captures its own client result
instead of inheriting shell state from an earlier case.

- RFC or draft: `Not applicable` — test lifecycle only; qlog protocol behavior is unchanged.

### Validation Cases

- `Not applicable` — test-runner lifecycle change; no endpoint case IDs were added.

### CONTRIBUTING.md

- Overall: `Pending`
- CI: `Pending` — required pull-request checks.

### DIFF
--- a/case_test/observability/qlog.sh
+++ b/case_test/observability/qlog.sh
@@ -18,10 +18,10 @@ rm -rf tp_localhost test_session xqc_token
 
 
 
+case_test_stop_server
 clear_log
 rm -rf tp_localhost test_session xqc_token
 echo -e "qlog disable ...\c"
-case_test_stop_server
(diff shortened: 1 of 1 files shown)


cherylsy (approved):
looks good

Yanmei-Liu:
PR #936 has been merged into main. Please rebase this branch onto the latest main before rerunning CI so it picks up the consolidated three-job workflow and does not run the retired standalone Harness Check / CodeQL jobs.

PR #946, opened 2026-09-10
[~] Fix #518: Prevent HQ request loops after delayed FIN

### LINKED ISSUE #518: [Bug]: xquic regularly and consistently fails Interop Runner runs

Yanmei-Liu:
### Mechanism

Return the actual copied HQ resource bytes and deliver the completed request only once. A CRLF-terminated request followed by a separate QUIC FIN previously corrupted read offsets and kept the server read callback looping over an already delivered path. Preserve CRLF completion while consuming a later FIN without replaying the request.

Regression coverage uses the existing CUnit and endpoint frameworks. CI reads the CUnit summary so diagnostic output cannot undercount passed tests.

- RFC: [9000 §2.2](https://www.rfc-editor.org/rfc/rfc9000.html#section-2.2) and [§19.8](https://www.rfc-editor.org/rfc/rfc9000.html#section-19.8).
- Reproduced from the [kwik multiplexing failure](https://interop.seemann.io/logs/quic/2026-09-09T18:34/xquic_kwik/multiplexing/). A fresh public Docker matrix run remains pending.

Fixes: #518

### Validation Cases

- 1702 — HQ request with inline FIN completes two transfers on one connection.
- 1703 — HQ request with FIN sent after its response completes a subsequent transfer on the same connection.

### CONTRIBUTING.md

- Overall: Pending
- Local regression: Complete
- CI: Pending — Build on Ubuntu; Test BabaSSL on Ubuntu 24.04 with GCC 13.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -305,24 +305,29 @@ jobs:
     - name: Check Test Result
       run: |
         log_file=xquic_test_ubuntu_24_babassl.log
-        unit_total=$(awk '/  Test: / {count++} END {print count + 0}' \
-          "$log_file")
-        unit_passed=$(awk '/  Test: .*passed/ {count++} END {print count + 0}' \
-          "$log_file")
-        unit_failed=$(awk '/  Test: .*FAILED/ {count++} END {print count + 0}' \
-          "$log_file")
+        unit_summary=$(awk '
+          $1 == "tests" && NF == 6 && $2 ~ /^[0-9]+$/ \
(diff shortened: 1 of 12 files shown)


dreamwind1985 (approved):
完善了测试用例，looks good

PR #960, opened 2026-09-17
[+] Fix #959: Bound the per-request HTTP/3 body buffer, with pause and re-arm

### LINKED ISSUE #959: [Bug]: HTTP/3 body_buf is unbounded when the application reads the body more slowly than the peer sends it
### What happened?

A request's `body_buf` holds HTTP/3 DATA payload that `xqc_h3_stream_process_data()` has read out of the transport stream and that the application has not yet collected with `xqc_h3_request_recv_body()`. At `2060a446273a6a8836da3b5e30c8498b11c100bb` no public setting names it -- `grep -rn body_buf include/` returns nothing. Its occupancy counter is moved, and the one place it is read tests it against zero:

```shell
$ grep -rn body_buf_count src/ include/ | sort
src/http3/xqc_h3_request.c:179:    h3_request->body_buf_count = 0;
src/http3/xqc_h3_request.c:759:            h3_request->body_buf_count--;
src/http3/xqc_h3_request.c:768:            h3_request->body_buf_count--;
src/http3/xqc_h3_request.c:785:    if (h3_request->body_buf_count == 0) {
src/http3/xqc_h3_request.h:39:    uint64_t                        body_buf_count;
src/http3/xqc_h3_stream.c:1127:                h3s->h3r->body_buf_count++;
```

One increment, and no comparison against a limit.

#### Expected behavior

A receiver whose application consumes a response body more slowly than the peer sends it holds bounded per-stream memory.

#### Actual behavior

`body_buf` grows for the length of the transfer.

#### Governing specification

- Document: RFC 9000
- Section: 4.1, Data Flow Control
- Link: https://www.rfc-editor.org/rfc/rfc9000.html#section-4.1
- Mechanism: stream-level flow control, MAX_STREAM_DATA.

### Steps To Reproduce

1. An HTTP/3 client and server on this repository, default `xqc_conn_settings_t`, one response body larger than the receiver will collect promptly.
2. On the receiving side keep calling `xqc_h3_request_recv_body()`, taking a small fixed amount per read callback.
3. Watch receiver RSS and `h3s->h3r->body_buf_count`; both grow for the length of the transfer.

Measured on a downstream deployment running a fork of this project, not on this repository and not in its CI: one request, 150 Mbit/s link, application reading at 20 Mbit/s, 60 s. `snd_bdy_sz:1073973516` at the sender against `rcvd_bdy_sz:152660474` at the receiver, receiver RSS +899 MB and still climbing. With that fork carrying this change and `max_body_buf_per_stream = 262144`, the same run: `snd_bdy_sz:162610475` against `rcvd_bdy_sz:152821110`, RSS +13 MB.

### Relevant log output

Not available.

mastercoding:
Fixes https://github.com/alibaba/xquic/issues/959

### Why

Without a bound the buffer grows for as long as the peer sends faster than the
application collects, and nothing slows the peer down: the `xqc_stream_recv()`
call that drives the flow-control update is made by the HTTP/3 layer as it
drains the transport stream, not by the application as it collects the body. So
the advertised limit follows arrival rather than consumption, and the
difference accumulates in `body_buf`.

Measured on a downstream deployment running a fork of this project, not on this
repository and not in its CI: one request, 150 Mbit/s link, application reading
at 20 Mbit/s, 60 s — the receiver's RSS grew 899 MB and was still climbing at
the end. With this change and `max_body_buf_per_stream = 262144`, the same run
grew 13 MB.

### Mechanism

`h3r->body_buf` holds HTTP/3 DATA payload that the HTTP/3 layer has read out of the transport stream and that the application has not yet collected. Nothing compares it to a limit: `grep -rn body_buf_count src/ include/` returns one increment, two decrements, an empty check and the declaration.

This adds `max_body_buf_per_stream` to `xqc_conn_settings_t`, appended last. Non-zero makes `xqc_h3_stream_process_data()` set a stream flag, call `xqc_stream_shutdown_read()` and break instead of making its next `xqc_stream_recv()` call, so `stream_data_in.next_read_offset` stops advancing; `xqc_h3_request_recv_body()` clears the flag and calls `xqc_stream_ready_to_read()` once the application has drained to a quarter of the limit. A node bound derived from the same setting covers DATA framed at a few bytes each. Default 0 leaves the read loop unchanged. No new frame, no negotiation, no wire change.

- RFC or draft: RFC 9000 Section 4.1

### Validation Cases

- The new coverage is ten cases in the unit suite. They do not build against the base, because the setting and the stream flag they reference do not exist there.

### CONTRIBUTING.md

- Overall: Pending
- Local regression: the unit suite passes in full; the endpoint cases were not run here.
- CI: Pending


mastercoding:
The two red Linux jobs fail on case tests that also fail on main, and this change doesn't run in them.

- **Test BabaSSL**: `MPNS_multipath_30_percent_loss_close_initial_path`, `MPNS_multipath_close_new_path` and `MPNS_multipath_30_percent_loss_close_new_path`. main failed the same three on d3b2a12 (https://github.com/alibaba/xquic/actions/runs/34442204659/job/102759413298) and on 278755e (https://github.com/alibaba/xquic/actions/runs/33036352734/job/98399825717). The first of them also failed on main in 5 other jobs since Aug 18. Locally, unmodified 2060a44 fails the same three in the same way (325 passed, 3 failed). Case 5 goes past its 120 s `--timeout`; the watchdog in `case_test/lib/common.sh` then kills the group's shared `test_server`, so cases 6 and 7 fail with "Connection refused".
- **Build on Ubuntu**: `h3_ext_bytestream_blocked_by_sndq_full`. main failed it on d3b2a12, in the BabaSSL job linked above. Locally on 2060a44, the case's own client command (`-x 309`) fails when the receiver reaches `XQC_MAX_STREAM_FRAME_BUFFERED_COUNT` and `xqc_insert_stream_frame` returns -613 (`XQC_ELIMIT`).
- The new pause and resume only run when `max_body_buf_per_stream > 0`. `test_client` and `test_server` leave it at 0. Under gcov, across 140 local replays, `limit > 0` never held and the pause/resume lines never ran.

Could a maintainer re-run the two failed jobs?

Yanmei-Liu (commented):
Thanks for the contribution and for the detailed CI investigation! The current Linux case-test failures appear unrelated to this change, although the required checks still need to be rerun successfully.

There are three edge cases that I think should be addressed in this PR because they are direct consequences of the newly introduced body-buffer limit and pause/resume behavior:

1. **Transport buffering while paused:** when the new code calls `xqc_stream_shutdown_read()`, the peer can still send within the already advertised flow-control window. Many small STREAM frames can then reach the existing frame-count limit and close the connection. A possible fix is to stop extending receive credit while the body is paused and ensure data already allowed by the current window can be stored without exhausting frame nodes, for example by coalescing adjacent fragments or using a byte-based transport budget. Credit can be extended again after the application drains below the low watermark.

2. **Scheduling after resume:** after `xqc_stream_ready_to_read()`, please schedule an engine wakeup, such as through `xqc_engine_wakeup_once()`, when the engine is not already running. This should avoid calling the main logic recursively while ensuring that consumption from an external application event can make progress without waiting for another packet.

3. **QPACK unblock replay:** please move the byte/node limit check into a shared DATA append helper used by both normal parsing and blocked-data replay. When the limit is reached, preserve the unconsumed portion in `blocked_buf`; after the application drains the body, resume that buffered input before reading more transport data.

